Issue #2, opened 2025-02-19
Name


ihlar:
We have a few options:

1. Keep TRAIN
2. Find something that fits TRONE
3. Completely new name, e.g., Throughput Advice Signaling Protocol (TASP)

I'd be fine with any of the above.

billwuqin:
Please implement the naming change in the draft.

Issue #1, opened 2025-02-19
Throughput advice format


ihlar:
While we're figuring out how to agree on a throughput advice format we could initially include both options:

- A TRAIN Packet always includes a payload field of size N after the SCID. 
- The third most significant bit is set to 1 when a network element updates the payload field. 
- If the bit is not set, the low 5 bits are used to signal a preset rate.

huitema:
We are back to the composability argument. TRAIN is designed to pick "the lowest of multiple assertions". During the discussion, I heard the view that there is no need for that, because in practice there is only one bottleneck and one TRAIN rewriter. I am really not sure about that -- I can really imagine a cascade of NATs, each believing that they are the only one enforcing bandwidth. The rule that you propose implies that any "rewriter" must check a specific bit, and if it is set jump to the payload and analyze it. Which implies that all on path agents will need to support "checking the payload", in which case I am not sure I see the point of having the information in the top bits.

ihlar:
I think it's fine if we eventually converge on a single format / location for the advice (I have my bias already).  
In order to get there it might be useful to have both options on the table so we can do some early implementations and measurements.

Also, with regards to composability. I agree that there are possible cases, but they are probably not that common. A network element that only considers the top bits could simply not send throughput advice if it sees the "check the payload" bit.

ihlar:
Based on the agreed design this issue is no longer relevant.

PR #3, opened 2025-02-26
Transparent Rate Optimization for Network Endpoints (TRONE)

ihlar:
An attempt at TRONE.
We can bikeshed the name in #2.

The PR uses TRAIN as a baseline so that it's easier to review the changes from TRAIN.
Note that this version changes the rate signal field to a field with explicit values, located after the CIDs.
If we're very unsure about the design tradeoffs we could consider doing both the TRAIN and TRONE ways as discussed in #1.

There's also a discussion on how network elements are capable of crafting TRONE packets.

### DIFF
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -264,7 +264,7 @@ Rate Signal {
 ~~~
 {: #fig-rate-signal title="Rate Signal Format"}
 
-Rate Limit is a 32 bit unsigned integer that indicates the maximum sustainable
+Rate Limit is a 32-bit unsigned integer that indicates the maximum sustainable
 throughput through the network element that sets it, expressed in Kilobits per
 second. A value of 0 indicates that there is no rate limit in place. A QUIC
 endpoint sets this value to 0 when sending a TRONE packet.
@@ -272,7 +272,7 @@ endpoint sets this value to 0 when sending a TRONE packet.
 Average Window, expressed in milliseconds is used to indicate the period over
 which a bitrate might be enforced.
 
-## Enpoint Processing of TRONE Packets
+## Endpoint Processing of TRONE Packets
 
 Processing a TRONE packet involves reading the value from the Rate Signal field.
 However, this value MUST NOT be used unless another packet from the same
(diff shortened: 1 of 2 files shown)


draft-thoji-scone-trone-protocol.md
```diff
@@ -309,58 +302,76 @@ transport parameters, and frame types registries established in {{Sections 22.2,
...
+  offset = 6 + dcid_len
+  scid_len = packet[offset]
+  offset = offset + 1 + scid_len
+
+  packet_rate = read_uint32(packet[offset : offset + 4])
+  packet_aw = read_uint32(packet[offset + 4 : offset + 8])
+  if packet_rate == 0 or target_rate < packet_rate:
+    write_uint32(packet[offset : offset + 4], target_rate)
+  if packet_aw == 0 or target_aw < packet_aw:
+    write_uint32(packet[offset + 4 : offset + 8], target_aw)
```
martinthomson, line 344: Even if I thought that this was necessary, this is probably not the right logic to use here.  You are going to independently reduce the rate and window size.  If you don't reduce one, but you reduce the other, that is sometimes going to reduce the actual send rate below your target.  That's another reason I don't like this design.

ihlar: Yeah, composability makes this solution a bit messy. I've been a bit back and forth on this.  My original thinking was to only check the rate and update the window unconditionally:
  if packet_rate == 0 or target_rate < packet_rate:
    write_uint32(packet[offset : offset + 4], target_rate)
    write_uint32(packet[offset + 4 : offset + 8], target_aw)
    
That could lead to the sending of bursts that are not tolerated by the element further upstream. 
It might be the least messy thing to do with this design though. (And I still think the case with chained elements who wish to signal rates is a rather exceptional one).

billwuqin: Thank for getting together baking this draft among authors, thanks for Martin to enage this discussion, I am wondering
Is there possible to leave this as one open issue or split it from PR#3 and move to new PR. 
I hope we can have one posted version before submission deadline, let us know what we can do to help move the way forward.

ihlar: The latest commit contains an update in response to Martin's comment, fixing the logic so that average window is not updated separately from updating the rate limit. Which is probably the right thing to do, given this design. 

There is a larger issue around the structure and placement of the rate signal, something the WG needs to resolve. I would hope that the merged document can contain the latest commit and then we have the broader discussion as an open issue, see #1 for instance.

martinthomson: To be clear, I did not approve this document.

PR #12, opened 2025-03-24
Initial attempt at adding version-dependent logarithmic rates

mjoras:
First pass, expecting there will be further changes needed. There seemed to be general consensus around this approach, which we believe we need to take before an adoption call.

### DIFF
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -252,7 +252,8 @@ defined by the TRONE protocol version.
 
 The Rate Signal field is set to 0x3F (63) when sent by a QUIC endpoint,
 indicating no rate limit is in place or that the TRONE protocol is not supported
-by network elements on the path.
+by network elements on the path. Vales 0x00 through 0x3E (62) represent the ceiling
+of rate being advised by the network element(s) on the path.
 
 For TRONE protocol version 0xTRONE1, the rate limits use a logarithmic scale with:
 * Base rate (b_min) = 100 Kbps
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -273,7 +273,7 @@ Some notable values in these ranges include:
 | Rate Signal | Version 0xTRONE1 Rate | Version 0xTRONE2 Rate |
 |:------------|:----------------------|:----------------------|
 | 0  | 100 Kbps | 10 Mbps |
-| 10 | 251 Kbps | 25.1 Mbps |
+| 10 | 316 Kbps | 31.6 Mbps |
 | 20 | 1 Mbps | 100 Mbps |
 | 30 | 3.16 Mbps | 316 Mbps |
 | 40 | 10 Mbps | 1 Gbps |
@@ -281,10 +281,6 @@ Some notable values in these ranges include:
 | 60 | 100 Mbps | 10 Gbps |
 | 62 | 126 Mbps | 12.6 Gbps |
 | 63 | No limit | No limit |
-
-All other values follow the logarithmic formula for each version. The
-overlapping ranges allow for finer granularity in the commonly used range of
-10-100 Mbps.
 
 ## Endpoint Processing of TRONE Packets
 
@@ -326,8 +322,8 @@ QUIC endpoints can enable the use of the TRONE protocol by sending TRONE packets
 A network element detects a TRONE packet by observing that a packet has a QUIC
 long header and one of the TRONE protocol versions (0xTRONE1 or 0xTRONE2).
 
-A network element then conditionally replaces the six bits of the Rate Signal
-field with a value of its choosing.
+A network element then conditionally replaces the Rate Signal field with
+values of its choosing.
 
 A network element might receive a packet that already includes a rate signal.
 The network element replaces the rate signal if it wishes to signal a lower
@@ -338,13 +334,10 @@ The following pseudocode indicates how a network element might detect a TRONE
 packet and replace an existing rate signal.
 
 ~~~ pseudocode
-target_rate_value = convert_rate_to_signal(target_rate, version)
-
 is_long = packet[0] & 0x80 == 0x80
-is_trone1 = compare(packet[1..5], TRONE1_VERSION)
-is_trone2 = compare(packet[1..5], TRONE2_VERSION)
-
-if is_long and (is_trone1 or is_trone2):
+packet_version = packet[1..5]
+if is_long and (packet_version == TRONE1_VERSION or packet_version == TRONE2_VERSION):
+  target_rate_value = convert_rate_to_signal(target_rate, packet_version)
   packet_rate_value = packet[0] & 0x3f
   if packet_rate_value == 0x3f or target_rate_value < packet_rate_value:
     packet[0] = packet[0] & 0xc0 | target_rate_value
@@ -356,18 +349,16 @@ network element signals a rate limit that falls within the range of both
 versions, it should signal appropriate values in each version. When the target
 rate limit is outside the range of a particular version, the network element
 should signal the minimum value (0) for rates below the range and preserve the
-"no limit" value (63) for rates above the range.
+original value for rates above the range.
 
 # Version Interaction {#version-interaction}
 
 The TRONE protocol defines two versions (0xTRONE1 and 0xTRONE2) that cover
 different but overlapping ranges of bitrates. This design allows for:
 
-1. Greater precision in commonly used ranges (particularly 10-100 Mbps) where
-   the ranges overlap
-2. Support for both very low bitrates (down to 100 Kbps) and very high bitrates
+1. Support for both very low bitrates (down to 100 Kbps) and very high bitrates
    (up to 12.5 Gbps)
-3. Graceful handling of network elements that might only recognize one version
+2. Graceful handling of network elements that might only recognize one version
    or some subset of future versions.
 
 ## Converting Between Versions
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -215,13 +215,12 @@ from {{Section 4 of INVARIANTS}}.
 TRONE Packet {
   Header Form (1) = 1,
   Reserved (1),
-  Reserved (6),
-  Version (32) = 0xTBD,
+  Rate Signal (6),
+  Version (32) = 0xTRONE1 or 0xTRONE2,
   Destination Connection ID Length (8),
   Destination Connection ID (0..2040),
   Source Connection ID Length (8),
   Source Connection ID (0..2040),
-  Rate Signal (TBD)
 }
 ~~~
 {: #fig-trone-packet title="TRONE Packet Format"}
@@ -230,7 +229,8 @@ The most significant bit (0x80) of the packet indicates that this is a QUIC long
 header packet.  The next bit (0x40) is reserved and can be set according to
 {{!QUIC-BIT=RFC9287}}.
 
-The low 6 bits (0x3f) of the first byte are reserved.
+The low 6 bits (0x3f) of the first byte contain the Rate Signal field. Values
+for this field are described in {{rate-signal}}.
 
 This packet includes a Destination Connection ID field that is set to the same
 value as other packets in the same datagram; see {{Section 12.2 of QUIC}}.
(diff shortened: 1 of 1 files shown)


draft-thoji-scone-trone-protocol.md
```diff
@@ -304,18 +338,78 @@ The following pseudocode indicates how a network element might detect a TRONE
...
+    packet[0] = packet[0] & 0xc0 | target_rate_value
 ~~~
 
+For TRONE packets with two versions, a network element applies the same rate
+policy to both, accounting for the different scales of each version. When a
+network element signals a rate limit that falls within the range of both
+versions, it should signal appropriate values in each version. When the target
+rate limit is outside the range of a particular version, the network element
+should signal the minimum value (0) for rates below the range and preserve the
+"no limit" value (63) for rates above the range.
```
kazuho, line 359:
```suggestion
should signal the minimum value (0) for rates below the range and preserve the
original value for rates above the range.
```
When there are multiple network elements, the Rate Signal field might convey a value other than 63.

draft-thoji-scone-trone-protocol.md
```diff
@@ -304,18 +338,78 @@ The following pseudocode indicates how a network element might detect a TRONE
...
+
+# Version Interaction {#version-interaction}
+
+The TRONE protocol defines two versions (0xTRONE1 and 0xTRONE2) that cover
+different but overlapping ranges of bitrates. This design allows for:
+
+1. Greater precision in commonly used ranges (particularly 10-100 Mbps) where
+   the ranges overlap
```
kazuho, line 367:
```suggestion
```

draft-thoji-scone-trone-protocol.md
```diff
@@ -244,13 +244,47 @@ TRONE packets SHOULD be included as the first packet in a datagram.  This is
...
 
-Specification of the the size and format of the Rate Signal field is left for
-future revisions of this document.
+The Rate Signal field in TRONE uses the low 6 bits (0x3f) of the first byte.
+This field is encoded as a logarithmically spaced distribution over a range
+defined by the TRONE protocol version.
+
+The Rate Signal field is set to 0x3F (63) when sent by a QUIC endpoint,
+indicating no rate limit is in place or that the TRONE protocol is not supported
+by network elements on the path.
```
kazuho, line 255: Do we want to state that values 0 to 62 represents the ceiling values?

kazuho (commented):
Thank you for the PR. Looks good overall.

draft-thoji-scone-trone-protocol.md
```diff
@@ -244,13 +244,47 @@ TRONE packets SHOULD be included as the first packet in a datagram.  This is
...
+| 30 | 3.16 Mbps | 316 Mbps |
+| 40 | 10 Mbps | 1 Gbps |
+| 50 | 31.6 Mbps | 3.16 Gbps |
+| 60 | 100 Mbps | 10 Gbps |
+| 62 | 126 Mbps | 12.6 Gbps |
+| 63 | No limit | No limit |
+
+All other values follow the logarithmic formula for each version. The
+overlapping ranges allow for finer granularity in the commonly used range of
+10-100 Mbps.
```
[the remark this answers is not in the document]
mjoras: You're right I got my wires crossed from my notes.

draft-thoji-scone-trone-protocol.md
```diff
@@ -324,13 +418,17 @@ makes when constructing datagrams. It is recommended that endpoints promptly
...
 them.
 
-An endpoint that has not recently sent a TRONE packet MAY treat receipt of one
-from its peer as a trigger to send a TRONE packet in the reverse direction. This
-way a peer that is receiving downlink data can influence the frequency of
-receiving rate signals.
+To support the full range of possible rate limits, endpoints SHOULD alternate
+between sending TRONE packets with version 0xTRONE1 and version 0xTRONE2. This
+allows network elements to signal rate limits across the entire supported range
+with appropriate granularity.
```
martinthomson, line 416: We should talk about the privacy trade-offs with respect to sending each.  I think that it is reasonable to send both unconditionally, but it might be better to only send these when the estimated bandwidth availability enters the range covered by each.

An application that is currently operating in the TRONE1 range could decide that it only needs to send TRONE1 packets.  That limits the number of markings that are needed and does not reveal anything extra.  The challenge being in ensuring that the network element is aware that TRONE markings are available.  For that, I would recommend always sending one packet at least once, even when the active usage is well outside of the protocol-defined ranges.


However, an application that MIGHT operate in just the TRONE2 range reveals information that the network does not otherwise obtain if it sends the TRONE2 signal.

draft-thoji-scone-trone-protocol.md
```diff
@@ -304,18 +338,78 @@ The following pseudocode indicates how a network element might detect a TRONE
...
-  offset = 6 + dcid_len
-  scid_len = packet[offset]
-  offset = offset + 1 + scid_len
-  packet_rate = read_rate_signal(packet[offset : offset + rate_signal_size])
-  if target_rate.is_lower_than(packet_rate):
-    write_rate_signal(packet[offset : offset + rate_signal_size], target_rate)
+is_trone1 = compare(packet[1..5], TRONE1_VERSION)
+is_trone2 = compare(packet[1..5], TRONE2_VERSION)
+
+if is_long and (is_trone1 or is_trone2):
```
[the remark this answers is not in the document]
martinthomson: Or more clever:

```python
i = find(versions, packet[1..5]) if packet[0] & 0x80 == 0x80 else None
if i is not None and ...
```

draft-thoji-scone-trone-protocol.md
mjoras: Agreed. Should I leave the text here while we work through this nuance or do you think there's something better to put in its place for now?

martinthomson: An issue for followup is fine.

draft-thoji-scone-trone-protocol.md
```diff
@@ -304,18 +338,78 @@ The following pseudocode indicates how a network element might detect a TRONE
...
+
+3. Network elements SHOULD apply appropriate rate signals to both versions when
+   they observe them, translating the target rate limit to the appropriate signal
+   value for each version's scale.
+
+## Example Conversions
+
+For a target rate limit of 30 Mbps:
+* In version 0xTRONE1, this corresponds to a signal value of approximately 49
+* In version 0xTRONE2, this corresponds to a signal value of approximately 9
```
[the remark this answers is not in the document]
kazuho: In the current state, I do not think we need to talk about rounding.

We state that the code points represent the ceiling. That means code points 0, 20, 40, 60 represent `<= 100Kbps`, `<= 1Mbps`, `<=10Mbps`, `<=100Mbps`, as they are divisible. OTOH, other values are indivisible. For example, 10<sup>1/20</sup> is 112201.84..., meaning that code point 1 represents bit rate up to 112,201bps.

Maybe all we need to provide is these examples.

Of course, we could go the other way and state that the 3 (or 4) most significant digits of the log function represents the bit rate...

draft-thoji-scone-trone-protocol.md
```diff
@@ -244,13 +244,44 @@ TRONE packets SHOULD be included as the first packet in a datagram.  This is
...
+|:------------|:----------------------|:----------------------|
+| 0  | 100 Kbps | 10 Mbps |
+| 10 | 316 Kbps | 31.6 Mbps |
+| 20 | 1 Mbps | 100 Mbps |
+| 30 | 3.16 Mbps | 316 Mbps |
+| 40 | 10 Mbps | 1 Gbps |
+| 50 | 31.6 Mbps | 3.16 Gbps |
+| 60 | 100 Mbps | 10 Gbps |
+| 62 | 126 Mbps | 12.6 Gbps |
+| 63 | No limit | No limit |
```
huitema, line 285: I think saying "63 = No limit" conflates two separate notions:

* the TRONE protocol is not supported by network elements on the path (definition of 0x3F above)
* the limit is higher that either 126Mbps or 12.6 Gbps.

We could remove that issue by defining the last lines of the table as:

| 60 | 100 Mbps | 10 Gbps |
| 62 | No limit | No limit |
| 63 |  TRONE not supported on path |  TRONE not supported on path |

draft-thoji-scone-trone-protocol.md
```diff
@@ -305,17 +336,72 @@ packet and replace an existing rate signal.
...
+
+# Version Interaction {#version-interaction}
+
+The TRONE protocol defines two versions (0xTRONE1 and 0xTRONE2) that cover
+different but overlapping ranges of bitrates. This design allows for:
+
+1. Support for both very low bitrates (down to 100 Kbps) and very high bitrates
+   (up to 12.5 Gbps)
+2. Graceful handling of network elements that might only recognize one version
+   or some subset of future versions.
```
huitema, line 363: Would be "up to 10Gbps" if we use only 62 actual ranges.

draft-thoji-scone-trone-protocol.md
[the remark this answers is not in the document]
huitema: I think it is fine to base the rounding on the table. I think we should say "pick the highest rate signal that correspond to a value lower or equal to the rate", but it might be a local decision.

huitema (commented):
I think this is fine, except for the confusion between "no TRONE on this path" and "no limit".

draft-thoji-scone-trone-protocol.md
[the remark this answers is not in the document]
martinthomson: We should offer advice to network operators on the consequences of different choices.  Rounding down has the effect of guaranteeing that the used capacity is lower than the target.  However, given that applications that employ rate adaptation will rarely be able to completely use the given bandwidth, it might be better to pick a higher number.  Especially given that any excess is never more than about 11% higher than the previous step.

draft-thoji-scone-trone-protocol.md
martinthomson: I'm not convinced that we need that feature.  What would  you do with that information?

huitema: I think there is a difference between "more than 10Gbps" and "don't know". In what case, the app has the go ahead to use a vast amount of bandwidth. In the other, the app has to proceed cautiously and rely on congestion control for a bandwidth estimate.

draft-thoji-scone-trone-protocol.md
martinthomson: `10M*10^{62/20}` = 12.589G; `10M*10^{61/20}` = 11.220G.  We only get 10G if we decide to use 61 ranges.

draft-thoji-scone-trone-protocol.md
martinthomson: I can't conceive of a scenario where an app would see an "uncapped" signal and decide that it can switch to sending at line rate.  You are taking the word of one path element and extrapolating to the entire path.

Let's say that you get a 1Mbps signal.  You are currently at ~30kbps.  It might be reasonable to skip a few steps in your capacity probing to get closer to 1Mbps.  But that might be based on an assumption about the relative availability of bandwidth in modern networks and maybe past experience with the same network; I doubt you would make the same guess at 1Tbps (if that were a possible signal, that is).

I conclude that while the congestion controller might be nudged a little, senders still need to have one and use that to probe the actual path availability.  The network still has queues and other users.  Expressions of policy limitations don't necessary entail guarantees, even at the node that is making that assertion.

draft-thoji-scone-trone-protocol.md
huitema: Sorry, I didn't do the math. But yes, that's the crux of the debate. We have 64 code points, from 0 to 63. If we reserve one to say "don't know", that means using only 62 code points, the last one marking the unbounded range.

draft-thoji-scone-trone-protocol.md
huitema: I really need to get going on writing the congestion control part. There are a couple of ways to use a "bandwidth cap". It could solve one specific problem, stop the "slow start" exponential growth early. It could fit with the "careful resume" mechanism -- very that the new connection has the same cap as the previous one before reusing values learned before. It could inform of bandwidth drop, controlling the building of queues when the congestion controller discovers the new capacity. It could inform "resume after congestion" mechanisms, to grow sending rate quicker after transient congestion stops. At the application level, it can inform choice of codecs...

draft-thoji-scone-trone-protocol.md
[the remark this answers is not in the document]
mjoras: File #17

draft-thoji-scone-trone-protocol.md
mjoras: Filed #18 for followup though I'm largely in agreement with @martinthomson on this.

mjoras:
Took all the suggestions and punted some things to issues. Are we okay with this now?

alvestrand:
Note the question in #14 - not an objection to merging this PR, but a heads-up that we may want to revisit this.

mjoras:
@huitema @kazuho @martinthomson any further thoughts or are we okay to merge this and iterate?

draft-thoji-scone-trone-protocol.md
huitema: We can come back to that after merging this PR.

huitema (approved):
Let's merge this. We can iterate later.

mjoras:
Who actually has merge approval perms? @britram could I get it provisionally just to h it the button? 🙂

draft-thoji-scone-trone-protocol.md
```diff
@@ -305,17 +336,72 @@ packet and replace an existing rate signal.
...
+packet_version = packet[1..5]
+if is_long and (packet_version == TRONE1_VERSION or packet_version == TRONE2_VERSION):
+  target_rate_value = convert_rate_to_signal(target_rate, packet_version)
+  packet_rate_value = packet[0] & 0x3f
+  if packet_rate_value == 0x3f or target_rate_value < packet_rate_value:
+    packet[0] = packet[0] & 0xc0 | target_rate_value
 ~~~
 
+For TRONE packets with two versions, a network element applies the same rate
+policy to both, accounting for the different scales of each version. When a
```
ihlar, line 348: This reads a bit like a single TRONE packet can have two versions, perhaps rewrite as:
``````suggestion
A network element applies the same rate policy to TRONE packets of both versions, accounting for the different scales of each version. When a
```

ihlar (approved):
Looks good to merge, one small editorial nit that we can fix now or later.

kazuho (approved):
+1 to merging.

PR #13, opened 2025-03-25
Description of DOS attack against intermediaries

### LINKED ISSUE #11: Spoofing attacks against intermediaries
In issue #9 we discuss the possibility of greasing TRONE. Greasing is cool for privacy, but the same mechanisms can be used as an attack. Suppose something like:

~~~
+------------+    +----------+       /---------------\       +-------------+
|end point A | -- | Trone NE | ---- ( Big bad Internet) ---- | end point B |
+------------+    +----------+       \---------------/       +-------------+
                                          |
                                          |
                                    +-------------+
                                    | wily hacker |
                                    +-------------+
~~~

The hacker creates Trone packets with:

* IP / UDP header: set IP source to address of B, IP destination to address of A
* SCONE packet: pick a destination CID and a destination CID
* 1RTT header: same destination  CID as SCONE packet
* 1 RTT content: enough random bytes to look plausible

The attacker needs to know at least a valid address for end point A, so the packets will be routed through the target NE, but after that there are many possibilities:

* an off path attacker can pick random values for the source address of "B" and for the CID.
* an on path attacker can pick pairs of addresses that match existing hosts "A" and "B" and either random CID or CID copied from observed traffic.

The possible effects would be:

* trick the NE into "remembering" lots of TRONE context, thus consuming lots of memory
* force the NE to process a vast quantity of TRONE packets, thus consuming lots of resource
* trick the NE into detecting a "systematic collision" (see issue #10) and thus disable TRONE for that flow.

huitema:
Adding a description of the attack listed in #11.

Close #11

### DIFF
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -426,7 +426,7 @@ and will have to process the packets. All the network elements between the injec
 point and the destination will have to process these packets.
 
 Attackers could send a high volume of these "fake" TRONE packets in
-a denial of service (DOS) attempt against intermediaries. The attack will
+a denial of service (DOS) attempt against network elements. The attack will
 force the intermediaries to process the fake packets. If intermediaries
 are keeping state for ongoing TRONE flows, the attack can cause the
 excessive allocation of memory resource. The mitigation there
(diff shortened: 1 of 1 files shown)


kazuho (approved):
Thank you for writing this!

Issue #8, opened 2025-03-21
Detail interaction with congestion control


huitema:
This came on repeatedly during the Bangkok IETF session. Congestion control and Trone are complementary, but we need to explain how that complementing works.

atiwariphd:
Totally agreed @huitema . We tried to describe the distinction between SCONE and ECN in an [earlier draft](https://datatracker.ietf.org/doc/draft-tomar-scone-ecn/) but I am glad you are owning up to describe the complementarity of Trone and congestion control wholistically.

martinthomson:
All of these seem to be variants on the scenario where known limits from the congestion controller is (much) lower than the throughput advice.  I think that we can do a few things:

Firstly, note that throughput advice is a cap, not a guarantee.

Then, note that throughput advice comes from a single network element, which cannot always make claims about other elements or links on the full end-to-end path.

Therefore, while it might suggest a target bitrate, there is no guarantee that that rate can be achieved.  This is weaker still than the careful resumption case, where at least you have recent evidence that a path between nominally the same endpoints supported a given throughput in the path.  Tweaking the probing of the congestion controller to move more aggressively toward the limit implied by throughput advice could be possible, but endpoints need to be especially sensitive to any congestion signals that might indicate a lower throughput in practice.

I fear that attempts to use this advice as a target rate will end up doing bad things if we aren't extremely careful in how we frame guidance.

huitema:
The prudent way to include TRONE advice is startup is indeed different from careful resume. Careful resume provides a target. TRONE provides an upper bound. Not the same, but valuable in its own right. Typical startup algorithm involve exponential increase and are probe to overshoot. Limiting startup to "probe up to this limit" would avoid overshooting, and the large queues that it causes.

huitema:
I was very specific. "When to stop startup" is a big issue. Typical startup process figures exponential growth, which means by the time the process has reached the desired capacity, there are twice too many packets flying. If you care about real time performance, that's not good. That does not mean the controller has to completely stop exploring. It could, but it would probably be wise to do that in a cautious way, maybe like the 2nd phase of Hystart++.

kazuho:
I agree about the problem. What I'm struggling to follow is how we could apply SCONE signals to Slow Start in case of real-time applications.

[...]

PR #16, opened 2025-03-26
First cut for congestion and TRONE interaction

huitema:
I think that we need to explicitly describe the interaction between congestion control and TRONE, if only to dispel any notion that TRONE replaces congestion control, as stated in issue #8. This is a first cut for describing such interactions.

### DIFF
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -372,6 +372,63 @@ TRONE packets could be stripped from datagrams in the network, which cannot be
 reliably detected.  This could result in a sender falsely believing that no
 network element applied a rate limit signal.
 
+## Interactions with congestion control
+
+TRONE and congestion control both provide the application with estimates
+of a path capacity. They are complementary. Congestion control algorithms
+are typically designed to quickly detect and react to congestion, i.e., to
+the "minimum" capacity of a path. TRONE informs the endpoint
+of the maximum capacity of a path.
+
+Consider for example a path in which the bottleneck router implements Early
+Congestion Notification as specified in the L4S architecture {{?RFC9330}}.
+If the path capacity diminishes, queues will build up and the router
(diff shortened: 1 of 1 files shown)


draft-thoji-scone-trone-protocol.md
```diff
@@ -372,6 +372,63 @@ TRONE packets could be stripped from datagrams in the network, which cannot be
...
+Consider for example a path in which the bottleneck router implements Early
+Congestion Notification as specified in the L4S architecture {{?RFC9330}}.
+If the path capacity diminishes, queues will build up and the router
+will immediately start increasing the rate at which packets are marked
+as "Congestion Experienced". The receiving endpoint will notice these marks,
+and inform its peer. The incoming congestion will be detected within
+1 round trip time (RTT). This scenario will play out whatever the reason
+for the change in capacity, whether due to increased competition between
+multiple applications or, for example, to a change in capacity of a wireless
+channel.
```
kazuho, line 392: I think text up to this paragraph if fine.

But for the remainder, I'm not sure if it'd be a good idea to imply TRONE as a signal that supplements congestion signals that indicate that more bandwidth might be available.

I wonder if it would be better to explain the difference of the time scales congestion signals and TRONE signals act upon.

The time scale of congestion signals is in the unit of RTT, while I'd presume TRONE signals to indicate average b/w available across ~10 seconds; see https://github.com/ietf-wg-scone/trone/issues/20. For video playback using chunked files, the two signals would be handled differently; congestion control applies to the transport, while the TRONE signal would be used for choosing the chunks with the appropriate bitrate.

huitema: Agreed. The following paragraphs are cut in the next commit.

kazuho (approved):
Thank you for the updates! LG

ihlar (approved):

Issue #5, opened 2025-03-20
Client-initiated signal


smishra1200:
1. Section 2 of the draft shows "QUIC Sender", ie, the server initiates the TRONE signal toward the client. This implies that TRONE signal is going to be server initiated. This will result in having the "network element" listens to all IP flows in the downlink direction. This is CPU intensive
2. The authors should consider a design that is independent of SNI-based flow detection
3. The authors should reconsider this design and allow for CLIENT initiated TRONE signaling (uplink direction) with implicit support for  flow detection

As a point of reference, section 6 of draft mishra-scone-usercase-00, added following requirement:
SCONE (aka TRONE) signal MUST be a client-application endpoint initiated to assist the network element (UPF/5G or PGW/4G) with the implicit flow detection
please see https://www.ietf.org/archive/id/draft-mishra-scone-usecase-00.html#section-6-1.1.1

ihlar:
TRONE packets are sent by each endpoint individually, so there is opportunity for the client to send an early TRONE packet such that a network element that keeps some flow state can use the presence of a TRONE packet as an indication to not do further DPI and SNI parsing etc. 
The problem in the current draft version is that endpoints need to indicate support using transport parameters before  sending TRONE packets. The authors have discussed an approach where an opportunistic TRONE "indication" can be appended to a QUIC initial packet. The network element can use this information as an indication that the client is willing to receive rate signals.

kazuho:
We have a lot of good discussion going on in #23, but maybe an issue is better to explain how people view the problem. So here are my thoughts:

As I understand, there are two types of flows with different characteristics:

1. flows rate-limited using CC
2. flows for which short bursts are allowed, with the assumption that the long-term bitrate is below the policy

Because SCONE signals are purely advisory that endpoints often cannot adhere to (see #29), network elements have to monitor the long-term bitrate of type 2 connections, and if they are exceeding the limit, demote them to type 1. This would be the ordinary business for any SCONE-compatible network elements with the capability to enforce bitrates.

If network elements want to demote flows without paying the cost of monitoring the long-term bitrate, they can consult the QUIC Version field of the first few packets sent by the server. As stated, there will be false positives; hence this is an optimization.

The question regarding client-driven indication is, IMO, if we want to have another knob for optimization, providing the capability to demote flows just by looking at the first packet from the client. Note we would have even more false positives than the first knob, because clients might be sending Indications to servers that do not support SCONE.

[...]

PR #23, opened 2025-04-15
Add "TRONE Indication" strawman.

mjoras:
As discussed in the meeting, and mentioned in #5, it is useful for an endpoint to signal its potential usage of TRONE to a network element. This initial pass uses a "TRONE Indication" packet which is probably excessive but is a fully reversed TRONE packet that's meant to go at the end of a datagram. This allows for the network element to first detect a QUIC initial, and then check the end of the packet for TRONE.

An alternative would be a more bespoke indicating like a concatenation of versions, or something like that.

### DIFF
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -310,7 +310,7 @@ This transport parameter is valid for QUIC versions 1 {{QUIC}} and 2
 transport parameters, and frame types registries established in {{Sections 22.2,
 22.3, and 22.4 of QUIC}}.
 
-# TRONE Indication
+# TRONE Indication {#trone-indication}
 
 QUIC endpoints can signal potential support for TRONE before the completion of
 the QUIC handshake by attaching a TRONE indication packet after a QUIC Initial
@@ -563,7 +563,7 @@ network element applied a rate limit signal.
 
 Network elements can detect potential TRONE support early in a connection by
 looking for TRONE indication packets attached to Initial packets, as described
-in {{TRONE Indication}}. This avoids the need for deep packet inspection of the
+in {{trone-indication}}. This avoids the need for deep packet inspection of the
 TLS handshake.
 
 
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -386,12 +386,12 @@ indication:
 is_quic = is_quic_datagram(datagram)
 if is_quic and datagram_length > MIN_INITIAL_SIZE:
   last_byte = datagram[datagram_length - 1]
-  
+
   if (last_byte & 0x80) == 0x80:
     version_start = datagram_length - 8
     version_end = datagram_length - 4
     potential_version = datagram[version_start:version_end]
-    
+
     if potential_version == TRONE1_VERSION or potential_version == TRONE2_VERSION:
       note_potential_trone_support(flow_tuple)
 ~~~
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -313,11 +313,16 @@ transport parameters, and frame types registries established in {{Sections 22.2,
 # TRONE Indication {#trone-indication}
 
 QUIC endpoints can signal potential support for TRONE before the completion of
-the QUIC handshake by attaching a TRONE indication packet after a QUIC Initial
-packet in the first UDP datagram. The TRONE indication provides an
+the QUIC handshake by attaching a TRONE indication packet after the last QUIC
+long header packet in the first UDP datagram. The TRONE indication provides an
 opportunistic signal to network elements that the client might support TRONE.
 Network elements can use this as an early hint, but must await confirmation of
 TRONE support by observing a full TRONE packet after the handshake completes.
+
+When sending QUIC short header packets in the first flight, endpoints SHOULD
+send them separately from the first datagram carrying the TRONE indication
+packet, to avoid receivers incapable of decoding TRONE packets from dropping
+the short header packet.
 
 ## TRONE Indication Packet
 
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -310,6 +310,104 @@ This transport parameter is valid for QUIC versions 1 {{QUIC}} and 2
 transport parameters, and frame types registries established in {{Sections 22.2,
 22.3, and 22.4 of QUIC}}.
 
+# TRONE Indication
+
+QUIC endpoints can signal potential support for TRONE before the completion of
+the QUIC handshake by attaching a TRONE indication packet after a QUIC Initial
+packet in the first UDP datagram. The TRONE indication provides an
+opportunistic signal to network elements that the client might support TRONE.
+Network elements can use this as an early hint, but must await confirmation of
+TRONE support by observing a full TRONE packet after the handshake completes.
+
+## TRONE Indication Packet
+
+A TRONE indication packet has a format with fully reversed field order compared
+to TRONE packets to facilitate detection by parsing from the end of the packet.
+
+{{fig-trone-indication-packet}} shows the format of the TRONE indication packet:
+
+~~~ artwork
+TRONE Indication Packet {
+  Source Connection ID (0..2040),
+  Source Connection ID Length (8),
+  Destination Connection ID (0..2040),
+  Destination Connection ID Length (8),
+  Version (32) = 0xTRONE1 or 0xTRONE2,
+  Rate Signal (6) = 0x3F,
+  Reserved (1),
+  Header Form (1) = 1
+}
+~~~
+{: #fig-trone-indication-packet title="TRONE Indication Packet Format"}
+
+A TRONE indication packet is attached directly after a QUIC Initial packet in
+the first UDP datagram. The Rate Signal field MUST be set to 0x3F (63) when
+sent as a TRONE indication packet.
+
+## Sending TRONE Indication
+
+A QUIC endpoint that supports TRONE MAY attach a TRONE indication packet after
+a QUIC Initial packet in the first UDP datagram. This is an opportunistic
+signal that is not guaranteed to be processed by the receiver, as the receiver
+has not yet confirmed its willingness to receive TRONE packets.
+
+When attaching a TRONE indication, the endpoint SHOULD:
+* Place it immediately after the QUIC Initial packet in the datagram
+* Construct the packet with reversed field order as shown in
+  {{fig-trone-indication-packet}}
+* Set the Source Connection ID and Destination Connection ID fields to match
+  those of the QUIC Initial packet
+* Set the Version field to either 0xTRONE1 or 0xTRONE2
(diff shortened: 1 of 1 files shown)


ihlar (approved):

draft-thoji-scone-trone-protocol.md
```diff
@@ -310,6 +310,104 @@ This transport parameter is valid for QUIC versions 1 {{QUIC}} and 2
 transport parameters, and frame types registries established in {{Sections 22.2,
 22.3, and 22.4 of QUIC}}.
 
+# TRONE Indication {#trone-indication}
+
+QUIC endpoints can signal potential support for TRONE before the completion of
+the QUIC handshake by attaching a TRONE indication packet after a QUIC Initial
+packet in the first UDP datagram. The TRONE indication provides an
```
kazuho, line 317:
```suggestion
the QUIC handshake by attaching a TRONE indication packet after the last QUIC long header
packet in the first UDP datagram. The TRONE indication provides an
```

Trone is designed to be QUIC-version-neutral, so it is better to avoid referring to concepts specific to QUIC v1. Also, even in v1, the first datagram that an endpoint sends might contain an Initial and a Handshake (consider servers).

draft-thoji-scone-trone-protocol.md
```diff
@@ -310,6 +310,104 @@ This transport parameter is valid for QUIC versions 1 {{QUIC}} and 2
...
+opportunistic signal to network elements that the client might support TRONE.
+Network elements can use this as an early hint, but must await confirmation of
+TRONE support by observing a full TRONE packet after the handshake completes.
```
kazuho, line 320:
```suggestion
TRONE support by observing a full TRONE packet after the handshake completes.

When sending QUIC short header packets in the first flight, endpoints SHOULD send them
separately from the first datagram carrying the TRONE indication packet, to avoid
receivers incapable of decoding TRONE packets from dropping the short header packet.
```

draft-thoji-scone-trone-protocol.md
```diff
@@ -310,6 +310,104 @@ This transport parameter is valid for QUIC versions 1 {{QUIC}} and 2
...
+{{fig-trone-indication-packet}} shows the format of the TRONE indication packet:
+
+~~~ artwork
+TRONE Indication Packet {
+  Source Connection ID (0..2040),
+  Source Connection ID Length (8),
+  Destination Connection ID (0..2040),
+  Destination Connection ID Length (8),
+  Version (32) = 0xTRONE1 or 0xTRONE2,
+  Rate Signal (6) = 0x3F,
```
kazuho, line 336: I think there is a question around what TRONE indication packets signal.

IIRC, TRONE is unidirectional.

Assuming the use of a TRONE indication packet signal that the sender is capable of _receiving_ TRONE packets, If so, I do not think need a rate signal here (and separately, I think we can use a different version number for indication packets).

kazuho (commented):
Thank you for the PR.

draft-thoji-scone-trone-protocol.md
mjoras: Agreed, I think all the fields here are likely excessive, but kept them in for now so we can discuss further. A new version is probably fine in terms of avoiding false positives.

mjoras:
@huitema @martinthomson thoughts on this?

We can merge as-is and remove fields later, or if we think it's important enough to cut them now I can change it to just be a QUIC version field.

kazuho:
I think we can merge this pull request (it'd be good to have all features being written down before the adoption call, but we can fix issues after adoption).

However, I realize that indications cannot be designed based solely on Invariants, because packet coalescing is a version-specific concept defined in RFC 9000. So the text has to reflect that at some point at least, and the fact might have an affect on what we append.

martinthomson:
I'm still unconvinced about the utility of this indication.  I know that we talked about it, but after thinking about it more, I don't think that it helps as much as you think it does.  Can we talk about this at the interim?

huitema:
Whatever we decide, we have to document what the concatenation of multiple Trone packets means, and whether it should be allowed. The draft assumes that a trone packet will be composed as:
~~~
TronePacket [*LongHeaderPacket] { LongHeaderPacket | ShortHeaderPacket }
~~~
This PR proposes to assign a meaning to:
~~~
LongHeaderPacket *LongHeaderPacket TronePacket 
~~~
I am not a great fan of that, and in particular not a great fan of encoding something at the tail of initial packets that must be padded to a fixed length. Implementation is a bit more complicated than it appears, and I am not really convince of the utility. But in any case, if we do that, we need to say something about potential combinations like:
~~~
TronePacket  InitialPacket TronePacket  
InitialPacket TronePacket _VX TronePacketVY
InitialPacket TronePacket  0RTTPacket
InitialPacket TronePacket  HandshakePacket
etc.
~~~

martinthomson:
If this is just a regular SCONE(TM) packet, then I would expect that it doesn't matter which one is used or where it appears.  At least to the network element.  But safe negotiation of QUIC depends on this being *after* QUIC packets.  Otherwise, a QUIC implementation that doesn't implement SCONE will be forced to throw out the rest of the datagram.  That's no good.

I'm more concerned that this doesn't help a network element much.  Even after this signal appears, the network element still doesn't know if the *server* is going to use the protocol.  So it still needs to put itself in a "SCONE-pending state", until it sees a SCONE(TM) packet.

kazuho:
@huitema Based on my understanding that we want to build SCONE (TRONE) on top of QUIC Invariants, I do not think think we should be considering how coalescing affects TRONE.

Invariants does not have the concept of coalesced packets.

If the version is TRONE, it is a TRONE packet. If it is not, then it is not.

If we are to have an indication, I think it has to be some junk being attached at the end of the datagram; please see https://github.com/mjoras/trone/pull/1. Regardless, it is not a coalesced packet.

huitema:
@kazuho I think that your [PR 1](https://github.com/mjoras/trone/pull/1) has the advantage of not muddying the concept of Trone packet, i.e., that they can only be found at the beginning of a datagram. We should say so.

ihlar:
> Quoting MT from [mjoras#1 (comment)](https://github.com/mjoras/trone/pull/1#discussion_r2053501602):
> 
> > (Indication at the end of the datagram) only works if all packets are understood enough to ignore them AND know their length. We have an existence proof of that not working in QUIC v1 (the short header packets) and there is no guarantee that other QUIC versions will retain a usable length.
> 
> I agree that this is the concern.

How much of a concern is this? 
Sending a TRONE indication packet is optional, a sender will know whether it is possible or not to append an indication. 
 
> If some feel strongly about having the indication, we could look for an alternative design.
> 
> One way forward would be to prepend TRONE Indication packets before QUIC v1 Initials (or any other version of QUIC that works end-to-end), and use SvcParamKey to advertise support of TRONE. If we take this path, the only case in which we would see an additional delay would be when the client does not have out-of-band knowledge and the name resolution uses A/AAAA records. How much do we need to care about such deployments?

This would be a cleaner solution though.

ihlar:
> I'm more concerned that this doesn't help a network element much. Even after this signal appears, the network element still doesn't know if the _server_ is going to use the protocol. So it still needs to put itself in a "SCONE-pending state", until it sees a SCONE(TM) packet.

You're right, this is all the network element knows until it sees a downlink SCONE packet.
However, current networks perform heavy DPI to determine whether flows are candidates for rate‑limiting policies. Early SCONE indications allow operators to omit that step.
Even in a “SCONE‑pending state,” a network element can initiate rate monitoring and potentially revert to throttling at some point, regardless of whether SCONE packets are observed in the downlink.

kazuho:
@ihlar 
> How much of a concern is this?
> Sending a TRONE indication packet is optional, a sender will know whether it is possible or not to append an indication.

I think the concern is that protocols other than QUIC v1 might not support TRONE, or that TRONE would become a constraint to the protocol designers, as it requires having a packet format that allows ignoring junk at the end. The strawman has issues even with QUIC v1 itself, it is ipossible to attach indications to path validations.

martinthomson:
>  Even in a “SCONE‑pending state,” a network element can initiate rate monitoring and potentially revert to throttling at some point, regardless of whether SCONE packets are observed in the downlink.

This seems to be key.  If you think that this state is distinct from a "SCONE-pending" state where you don't get an indicator, why would you not do that for all QUIC flows once you hit a certain adoption threshold?

ihlar:
> I think the concern is that protocols other than QUIC v1 might not support TRONE, or that TRONE would become a constraint to the protocol designers, as it requires having a packet format that allows ignoring junk at the end. 

Makes sense, thanks.

> The strawman has issues even with QUIC v1 itself, QUIC v1 does not allow indications to be appended to packets sent on new paths after the handshake (as they would be short header packets).

oth we should be able to send "real" SCONE packets at that point though?

kazuho:
>> The strawman has issues even with QUIC v1 itself, QUIC v1 does not allow indications to be appended to packets sent on new paths after the handshake (as they would be short header packets).
> 
> oth we should be able to send "real" SCONE packets at that point though?

Not really, because SCONE packets and Indication are sent in different directions. The Indication as defined by this PR has only one form: appended at the end of the datagram, which is incompatible with short header packets of QUIC v1.

But separately, I think @martinthomson is raising an interesting question.

draft-thoji-scone-trone-protocol.md
```diff
@@ -310,6 +310,109 @@ This transport parameter is valid for QUIC versions 1 {{QUIC}} and 2
...
+
+## Sending TRONE Indication
+
+A QUIC endpoint that supports TRONE MAY attach a TRONE indication packet after
+a QUIC Initial packet in the first UDP datagram. This is an opportunistic
+signal that is not guaranteed to be processed by the receiver, as the receiver
+has not yet confirmed its willingness to receive TRONE packets.
+
+When attaching a TRONE indication, the endpoint SHOULD:
+* Place it immediately after the QUIC Initial packet in the datagram
```
martinduke, line 360: Is this consistent with a Datagram that has a coalesced 0RTT packet, or Mozilla-like zero padding? The parsing rules indicate this should be the very last thing.

draft-thoji-scone-trone-protocol.md
```diff
@@ -310,6 +310,109 @@ This transport parameter is valid for QUIC versions 1 {{QUIC}} and 2
...
+    if potential_version == TRONE1_VERSION or potential_version == TRONE2_VERSION:
+      note_potential_trone_support(flow_tuple)
+~~~
+
+This approach allows network elements to detect TRONE support without needing
+to parse the full QUIC Initial packet structure or perform deep packet
+inspection of the TLS handshake.
+
+Network elements that observe TRONE indication packets MAY:
+* Note the UDP 4-tuple for potential future TRONE handling
```
martinduke, line 409: It would be helpful for understanding to elaborate more on the implications of these actions? An operator can do whatever it wants, but it is not clear to me what actions we expect them to take in terms of treatment of the flow (besides not trying to extract the SNI)

DavidSchinazi, line 409: Can you elaborate on how the network elements reacts to this indication? The network element needs to wait for a TRONE signal in the server-to-client direction in order to confirm that the flow supports TRONE. Why is this better than just having the network element look at the first packet in the server-to-client direction of each 5-tuple?

Issue #19, opened 2025-03-27
Privacy considerations when sending different TRONE protocol versions


mjoras:
See discussion on #12 

The draft currently recommends alternating between TRONE1 and TRONE2 versions without addressing the privacy implications. We should think about this more and at least add some considerations to the text.

To ensure network elements know TRONE is supported while minimizing information disclosure, one idea would be:

- Send at least one packet of each version initially even if the application only operates in one range.
- Subsequently prioritize sending the version most relevant to current bandwidth usage

ihlar:
I think it would be fine to keep alternating the versions at all times, just use two datagrams to send TRONE probes. I suppose that would be the best solution from a privacy perspective (unless I'm missing something). The additional cost of alternating should be low, and there are other potential benefits as discussed in #10 .

kazuho:
To be precise, will it be:
a) for the first burst send TRONE1 and for the next burst send TRONE 2, or
b) for each burst send both TRONE 1 and TRONE2?

I assume it'd be _b", but trying to confirm.

Regarding the overhead, I agree that it'd be negligible. The lowest bitrate being proposed is 100Kbps which is ~10pps assuming 1200 byte packets. That'd mean that, if endpoints are to receive updated TRONE signals one per every 10 seconds (see #20), the difference is sending one or two QUIC packets with TRONE headers for every 100 UDP datagrams.

ihlar:
Yes agree that it should be _b_.

kazuho:
As stated in https://github.com/ietf-wg-scone/trone/issues/14#issuecomment-2809099526, we could allow network elements rewrite TRONE2 packets (carrying higher bit rates) to TRONE1.

ihlar:
Closed by #25

PR #25, opened 2025-04-17
allow network elements rewrite the version field

### LINKED ISSUE #14: Is the "version=range" paradigm appropriate?
As presented at IETF 122, we define a "version" field that is used to choose between possible ranges.
This seems strange both from a cleanliness standpoint and from a greasing standpoint.

When there are only two versions, the version field amounts to an extra bit in the range specifier - by including it directly in the range specifier (making it 0-127 rather than 0-63), we gain an unified range with a larger span (or a finer resolution, if we choose to reduce the multiplier in the exponential spacing).

If we desire to ensure that we can introduce more versions of the TRONE signal later, two versions with near-identical semantics are not enough.

kazuho:
As discussed in https://github.com/ietf-wg-scone/trone/issues/14#issuecomment-2809099526 and during yesterday's interim, we can simplify the design by allowing network elements rewrite the version field.

Now, senders need to send only one packet (0xTRONE2 with rate=0x3f) instead of sending one for each of the two versions, and receivers no longer need to calculate the minimum of two values.

Maybe closes #14 and #19.

### DIFF
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -266,7 +266,7 @@ For TRONE protocol version 0xTRONE2, the rate limits use a logarithmic scale wit
 
 * Bitrate at value n = b_min * 10^((n + 64)/20)
 
-With two versions combined, bitrates between 100 Kbps and 178 Gbps can be
+With two versions combined, bitrates between 100 Kbps and 199.5 Gbps can be
 expressed.
 
 Some notable values in these ranges include:
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -286,7 +286,7 @@ Some notable values in these ranges include:
 | 0xTRONE2 | 36          | 10 Gbps   |
 | 0xTRONE2 | 46          | 31.6 Gbps |
 | 0xTRONE2 | 56          | 100 Gbps  |
-| 0xTRONE2 | 62          | 178 Gbps  |
+| 0xTRONE2 | 62          | 199.5 Gbps  |
 | 0xTRONE2 | 63          | No limit  |
 
 ## Endpoint Processing of TRONE Packets
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -253,8 +253,8 @@ defined by the TRONE protocol version.
 When sent by a QUIC endpoint, the Version field of a TRONE packet is set to
 0xTRONE2 and the Rate Signal field is set to 0x3F (63), indicating no rate limit
 is in place or that the TRONE protocol is not supported by network elements on
-the path. All other values (0x00 through 0x3F of protocol version 0xTRONE1 and
-0x00 through 0x3E of protocol version 0xTRONE2) represent the ceiling of rates
+the path. All other values (0x00 through 0x3F for protocol version 0xTRONE1 and
+0x00 through 0x3E for protocol version 0xTRONE2) represent the ceiling of rates
 being advised by the network element(s) on the path.
 
 For TRONE protocol version 0xTRONE1, the rate limits use a logarithmic scale with:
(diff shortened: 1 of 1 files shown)


draft-thoji-scone-trone-protocol.md
```diff
@@ -250,38 +250,44 @@ The Rate Signal field in TRONE uses the low 6 bits (0x3f) of the first byte.
...
-* Base rate (b_min) = 10000 Kbps (10 Mbps)
-* Maximum rate ≈ 12.5 Gbps
-* Bitrate at value n = b_min * 10^(n/20)
 
-This creates two overlapping ranges:
-* Version 0xTRONE1: 100 Kbps to 125 Mbps
-* Version 0xTRONE2: 10 Mbps to 12.5 Gbps
+* Bitrate at value n = b_min * 10^((n + 64)/20)
+
+With two versions combined, bitrates between 100 Kbps and 178 Gbps can be
```
ihlar, line 269: 178 is at n = 61 
n = 62 should be ~199.53. 

```suggestion
With two versions combined, bitrates between 100 Kbps and 199.5 Gbps can be
```

draft-thoji-scone-trone-protocol.md
```diff
@@ -250,38 +250,44 @@ The Rate Signal field in TRONE uses the low 6 bits (0x3f) of the first byte.
...
+| 0xTRONE1 | 40          | 10 Mbps   |
+| 0xTRONE1 | 50          | 31.6 Mbps |
+| 0xTRONE1 | 60          | 100 Mbps  |
+| 0xTRONE2 | 6           | 316 Mbps  |
+| 0xTRONE2 | 16          | 1 Gbps    |
+| 0xTRONE2 | 26          | 3.16 Gbps |
+| 0xTRONE2 | 36          | 10 Gbps   |
+| 0xTRONE2 | 46          | 31.6 Gbps |
+| 0xTRONE2 | 56          | 100 Gbps  |
+| 0xTRONE2 | 62          | 178 Gbps  |
```
ihlar, line 289:
```suggestion
| 0xTRONE2 | 62          | 199.5 Gbps  |
```

draft-thoji-scone-trone-protocol.md
```diff
@@ -338,68 +344,23 @@ packet and replace an existing rate signal.
...
-The formulas for conversion are:
-
-* For version 0xTRONE1: bitrate = 100 Kbps * 10^(signal/20)
-* For version 0xTRONE2: bitrate = 10 Mbps * 10^(signal/20)
-
-## Handling Multiple Versions
-
-Endpoints should observe the following guidelines when sending and processing
-TRONE packets:
+   (up to 178 Gbps)
```
ihlar, line 361:
```suggestion
   (up to 199.5 Gbps)
```

ihlar (commented):
Thanks for this PR. 
Looks really good, small nit commented below.

draft-thoji-scone-trone-protocol.md
kazuho: Thank you!

mjoras:
Thanks for this! Was writing up something basically the same.

Do we think we need any language stipulating that network elements should explicitly ignore any version that is not 0xTRONE1/2? It is sort of implied by QUIC long header semantics, but it might be worth noting explicitly that an element shouldn't stomp on a potential protocol it doesn't understand.

kazuho:
Thank you for checking!

> Do we think we need any language stipulating that network elements should explicitly ignore any version that is not 0xTRONE1/2? It is sort of implied by QUIC long header semantics, but it might be worth noting explicitly that an element shouldn't stomp on a potential protocol it doesn't understand.

Maybe raise as a different issue?

But FWIW, I do not think we want to neither allow or prohibit endpoints from mangling other versions of QUIC. It is not our business, and in the future, we might see another version of QUIC that recommends endpoints doing such a thing (example: a super-ECN signal that changes the bits in UDP payload). I think what we might need to add is a pointer to RFC 8999 regarding the topic.

ihlar:
> > Do we think we need any language stipulating that network elements should explicitly ignore any version that is not 0xTRONE1/2? It is sort of implied by QUIC long header semantics, but it might be worth noting explicitly that an element shouldn't stomp on a potential protocol it doesn't understand.
> 
> Maybe raise as a different issue?

Agree to handling this in a separate issue. The way the design is described is good enough for a document to get adopted. We can decide on additional guidance later on.

ihlar (approved):

ihlar:
I think we should merge this prior to an adoption call. Any concerns with that?

mjoras:
Merge it!

PR #28, opened 2025-04-28
trone -> scone

ihlar:
As agreed at interim and confirmed with authors, let's use SCONE as the protocol name.


martinthomson (approved):
Should we ask the chairs to move the repo first, then we can ensure that the links don't need to update twice.

huitema:
+1 @martinthomson. It would be natural to make that PR post adoption as a WG item, move from `draft-thoji-scone-trone-protocol-nn` to `draft-ietf-scone-protocol-00`, and have the new draft replace the old one.

billwuqin:
To move repo, we have two options:
1. rename repo
2. create a new repo and delete old "trone" repo
I would suggest to take option 1, since option 1 leaves freedom to change draft name by yourselves. If we take option 2, all PRs and open issues might be lost. 
Make sense?

ihlar:
> To move repo, we have two options:
> 
> 1. rename repo
> 2. create a new repo and delete old "trone" repo
>    I would suggest to take option 1, since option 1 leaves freedom to change draft name by yourselves. If we take option 2, all PRs and open issues might be lost.
>    Make sense?

Agree, option 1 sounds best.

billwuqin:
Okay, I have renamed repo as 'scone' repo. Hope everything works well.

martinthomson:
Given where things stand, I'm OK with merging this now.  @huitema suggested that our (prospective) WG-00 not include any text changes, but I don't think we need to worry about a simple s/trone/scone/i for that; we have to rename other things when submitting a -00 anyway.

PR #30, opened 2025-05-01
When detecting flows not adhering to SCONE signals, RECOMMEND acting as if SCONE was not used

### LINKED ISSUE #29: How hard can network elements throttle connections that seemingly abuse SCONE?
Some of the interesting comments last meeting I've heard are:
* Network elements can track usage per user and punish users or connections of such users that seemingly abuse SCONE, more so than users / connections that do not use SCONE at all.
* Web browsers might expose SCONE rate signals to JavaScript so that JavaScript can make the decision regarding what to fetch next.

I am concerned if we can allow both at the same time. I doubt that, if web browsers are going to expose SCONE rate signals to JavaScript, then there'd be a non-negligible probability of misuse (either intentional or accidental), and that we'd not want to punish web browsers for such mistakes.

kazuho:
This pull request expands what SCONE being advisory means to network elements (i.e., that they should handle non-adhering SCONE flows as if it was not a SCONE flow).

Closes #29.

### DIFF
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -190,9 +190,10 @@ The fact that an endpoint requests bitrate signals does not necessarily mean
 that it will adhere to them; in some cases, the endpoint cannot. For
 example, a flow may initially be used to serve video chunks, with the client
 selecting appropriate chunks based on bitrate signals, but later switch to a
-bulk download for which bitrate adaptation is not applicable. In other cases,
-the flow may be a tunnel carrying multiple connections, some of which may not be
-capable of handling SCONE signals. Therefore, when a network element detects a
+bulk download for which bitrate adaptation is not applicable. Composite flows
+from multiple applications, such as tunneled flows, might only have a subset of
+the involved applications that are capable of handling SCONE signals. Therefore,
+when a network element detects a
 flow using more bandwidth than advertised via SCONE, it SHOULD fall back to rate
 limiting based on congestion control signals, as if the flow were not using
 SCONE at all.
--- a/draft-thoji-scone-trone-protocol.md
+++ b/draft-thoji-scone-trone-protocol.md
@@ -186,9 +186,16 @@ signaling occurs as this is specific to the application in use.
 A signal does not prove that a higher rate would not be successful.  Endpoints
 that receive this signal therefore need to treat the information as advisory.
 
-As an advisory signal, network elements cannot assume that endpoints will
-respect the signal.  Though this might reduce the need for more active rate
-limiting, how rate limit enforcement is applied is a matter for network policy.
+The fact that an endpoint requests bitrate signals does not necessarily mean
+that it will adhere to them; in some cases, the endpoint cannot. In certain
+situations, a flow may initially be used to serve video chunks, with the client
+selecting appropriate chunks based on bitrate signals, but later switch to a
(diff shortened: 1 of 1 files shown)


draft-thoji-scone-trone-protocol.md
```diff
@@ -186,9 +186,16 @@ signaling occurs as this is specific to the application in use.
 A signal does not prove that a higher rate would not be successful.  Endpoints
 that receive this signal therefore need to treat the information as advisory.
 
-As an advisory signal, network elements cannot assume that endpoints will
-respect the signal.  Though this might reduce the need for more active rate
-limiting, how rate limit enforcement is applied is a matter for network policy.
+The fact that an endpoint requests bitrate signals does not necessarily mean
+that it will adhere to them; in some cases, the endpoint cannot. In certain
+situations, a flow may initially be used to serve video chunks, with the client
```
martinthomson, line 191:
```suggestion
that it will adhere to them; in some cases, the endpoint cannot. For
example, a flow may initially be used to serve video chunks, with the client
```

draft-thoji-scone-trone-protocol.md
```diff
@@ -186,9 +186,16 @@ signaling occurs as this is specific to the application in use.
...
+selecting appropriate chunks based on bitrate signals, but later switch to a
+bulk download for which bitrate adaptation is not applicable. In other cases,
+the flow may be a tunnel carrying multiple connections, some of which may not be
+capable of handling SCONE signals. Therefore, when a network element detects a
```
martinthomson, line 195:
```suggestion
bulk download for which bitrate adaptation is not applicable. Composite flows
from multiple applications, such as tunneled flows, might only have a subset of
the involved applications that are capable of handling SCONE signals. Therefore,
when a network element detects a
```

draft-thoji-scone-trone-protocol.md
```diff
@@ -186,9 +186,16 @@ signaling occurs as this is specific to the application in use.
...
+flow using more bandwidth than advertised via SCONE, it SHOULD fall back to rate
+limiting based on congestion control signals, as if the flow were not using
+SCONE at all.
```
martinthomson, line 198:
```suggestion
flow using more bandwidth than advertised via SCONE, congestion control
signals might be necessary to ensure that the flow does not exceed the
limits that are set in policy.
```

martinthomson (approved):
Some suggestions, but thanks for doing the work here.

draft-thoji-scone-trone-protocol.md
kazuho: Thank you for the review. Other changes made total sense, but I'm not fully certain about this one.

While it is true that CC might be necessary to force flows to be in policy, the proposed change does not address the original issue; i.e., that network elements should not punish non-adhering flows more so than non-SCONE flows. If network elements start behaving as such, endpoint developers would have no choice but to disable SCONE.

What about something like: "congestion control signals can be used to ensure that the flow performs on par with non-SCONE flows."?

kazuho: @martinthomson Went ahead with 387b028, please let me know what you think.

martinthomson (approved):
That's better than what I suggested, thanks @kazuho.

huitema (approved):
No objection.

ihlar (approved):
Looks good, thanks.

PR #31, opened 2025-05-12
Add usage of SCONE for throughput advice outside of rate-limiting

Kevsy:
This PR explicitly includes in scope networks that do not rate limit, but wish to advise on network conditions to help achieve the optimal sender rate. 

Notes on changes:
Throughout: replaced 'rate limit signal' with 'throughput advice signal' to account for both uses of SCONE (to signal rate limits/to signal throughput advice based on current network conditions)

Line 68: Changed 'Many access networks limit the maximum data rate that attached devices are able to attain'  (because all shared access networks do this due to finite bandwidth) to 'Many access networks apply rate limits to constrain the data rate of attached devices.' (to more explicitly indicate active rate limiting).

Line 164: Removed "In that case, capacity is less likely to be split between multiple active flows." because capacity will always be divided between multiple active flows (otherwise they would not be active)

Line 165: Added that a SCONE signal based on the network's calculation of throughput advice would take into account congestion, whereas (as stated in the original) one simply stating a rate limit would not.

(Other suggested changes are editorial)

### DIFF
--- a/draft-thoji-scone-protocol.md
+++ b/draft-thoji-scone-protocol.md
@@ -136,7 +136,7 @@ strictly advisory.
 
 ## Independent of Congestion Signals
 
-throughput advice signals are not a substitute for congestion feedback.  Congestion
+Throughput advice signals are not a substitute for congestion feedback.  Congestion
 signals, such as acknowledgments, provide information on loss, delay, or ECN
 markings {{?ECN=RFC3168}} that suggest the real-time condition of a network
 path.  Congestion signals might suggest a throughput that is different from the
--- a/draft-thoji-scone-protocol.md
+++ b/draft-thoji-scone-protocol.md
@@ -166,7 +166,9 @@ expected throughput.  That throughput will only be achievable when there is no
 significant data flowing in the same scope.  In the presence of other flows,
 congestion limits are likely to determine actual throughput.
 
-
+This makes the application of signals most usefully applied to a downlink flow
+in access networks, close to an endpoint. In that case, capacity is less likely
+to be split between multiple active flows.
 
 ## Per-Flow Signal
 
--- a/draft-thoji-scone-protocol.md
+++ b/draft-thoji-scone-protocol.md
@@ -14,6 +14,7 @@ workgroup: SCONE
 keyword:
  - locomotive
  - pastry
+ - clotted cream THEN jam
 venue:
   group: "SCONE"
   type: "Working Group"
@@ -57,23 +58,33 @@ informative:
 --- abstract
 
 On-path network elements can sometimes be configured to apply rate limits to
-flows that pass them.  This document describes a method for signaling to
-endpoints that rate limiting policies are in force and  what that rate limit is.
+flows that pass them. Access networks and their attached devices may also
+experience volatile conditions that affect the sustainable network throughput. 
+
+This document describes a method for signaling to
+endpoints that rate limiting policies are in force and what that rate limit is,
+and for non-rate limiting networks to signal throughput advice based on network
+conditions.
 
 
 --- middle
 
 # Introduction
 
-Many access networks limit the maximum data rate that attached devices are able
-to attain.  This is often done without any indication to the applications
+Many access networks apply rate limits to constrain the data rate of attached
+devices. This is often done without any indication to the applications
 running on devices.  The result can be that application performance is degraded,
 as the manner in which rate limits are enforced can be incompatible with the
 rate estimation or congestion control algorithms used at endpoints.
 
 Having the network indicate what its rate limiting policy is, in a way that is
-accessible to endpoints, might allow applications to use this information when
+accessible to endpoints, allows applications to use this information when
 adapting their send rate.
+
+In addition, access networks without rate limiting policies may also provide 
+information to endpoints that can help account for connection state, such as 
+network congestion, volatile signal strength, or radio bearers with constrained
+bandwidth.
 
 The Standard Communication with Network Elements (SCONE) protocol is
 negotiated by QUIC endpoints.  This protocol provides a means for network
@@ -122,19 +133,19 @@ This protocol only works for flows that use the SCONE packet ({{packet}}).
 
 The protocol requires that packets are modified as they transit a
 network element, which provides endpoints strong evidence that the network
-element has the power to drop packets; though see {{security}} for potential
-limitations on this.
+element has the power to apply rate limiting; though see {{security}} for
+potential limitations on this.
 
-The rate limit signal that this protocol carries is independent of congestion
+The throughput advice signal that this protocol carries is independent of congestion
 signals, limited to a single path and UDP packet flow, unidirectional, and
 strictly advisory.
 
 ## Independent of Congestion Signals
 
-Rate limit signals are not a substitute for congestion feedback.  Congestion
+throughput advice signals are not a substitute for congestion feedback.  Congestion
 signals, such as acknowledgments, provide information on loss, delay, or ECN
-markings {{?ECN=RFC3168}} that indicate the real-time condition of a network
-path.  Congestion signals might indicate a throughput that is different from the
+markings {{?ECN=RFC3168}} that suggest the real-time condition of a network
+path.  Congestion signals might suggest a throughput that is different from the
 signaled rate limit.
 
 Endpoints cannot assume that a signaled rate limit is achievable if congestion
@@ -145,22 +156,22 @@ congestion controller.
 
 ## Unspecified Scope
 
-Modifying a packet does not prove that the rate limit that is indicated would be
+Modifying a packet does not prove that the throughput that is indicated would be
 achievable.  A signal that is sent for a specific flow is likely enforced at a
 different scope.  The extent of that scope is not carried in the signal.
 
 For instance, limits might apply at a network subscription level, such
 that multiple flows receive the same signal.
 
-Endpoints can therefore be more confident in the rate limit signal as an
+Endpoints can therefore be more confident in the throughput signal as an
 indication of the maximum achievable throughput than as any indication of
-expected throughput.  That throughput will only be achievable when there is no
-significant data flowing in the same scope.  In the presence of other flows,
-congestion limits are likely to determine actual throughput.
+expected throughput.  Where the throughput signal simply represents a rate
+limit, then the advised throughput will only be achievable on an uncongested network.
+Whereas a throughput signal calculated by a non rate-limiting network should be 
+based on current network and attached device signal conditions, and hence account for
+any network congestion.
 
-This makes the application of signals most usefully applied to a downlink flow
-in access networks, close to an endpoint. In that case, capacity is less likely
-to be split between multiple active flows.
+
 
 ## Per-Flow Signal
 
@@ -172,9 +183,9 @@ involved.
 
 ## Undirectional Signal
 
-The endpoint that receives a rate limit signal is not the endpoint that might
+The endpoint that receives a throughput advice signal is not the endpoint that might
 adapt its sending behavior as a result of receiving the signal.  This ensures
-that the rate limit signal is attached to the flow that it is mostly likely to
+that the throughput advice signal is attached to the flow that it is mostly likely to
 apply to.
 
 An endpoint might need to communicate the value it receives to its peer in order
@@ -197,12 +208,12 @@ when a network element detects a flow using more bandwidth than advertised via
 SCONE, it might switch to applying its policies for non-SCONE flows, using
 congestion control signals.
 
-The time and scope over which a rate limit applies is not specified.  The
-effective rate limit might change without being signaled.  The signaled limit
-can be assumed to apply to the flow of packets on the same UDP address tuple for
-the duration of that flow.  Rate limiting policies often apply on the level of a
-device or subscription, but endpoints cannot assume that this is the case.  A
-separate signal can be sent for each flow.
+The time and scope over which throughput advice applies is not specified.  The
+effective throughput advice might change without being signaled.  The signaled
+advice can be assumed to apply to the flow of packets on the same UDP address 
+tuple for the duration of that flow.  For rate limiting networks, rate limiting
+policies often apply on the level of a device or subscription, but endpoints
+cannot assume that this is the case.  A separate signal can be sent for each flow.
 
 
 # Conventions and Definitions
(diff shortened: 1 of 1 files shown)


draft-thoji-scone-protocol.md
```diff
@@ -57,24 +57,34 @@ informative:
 --- abstract
 
 On-path network elements can sometimes be configured to apply rate limits to
```
martinthomson, line 59:
```suggestion
This document describes a protocol where on-path network elements
can give endpoints their perspective on what the maximum achievable
throughput might be for QUIC flows.
```

This can be shorter, and I think we don't need to get into what might cause a network element to offer this advice, rather we just need to establish that network elements have a perspective and that that perspective might be useful to communicate.

draft-thoji-scone-protocol.md
```diff
@@ -57,24 +57,34 @@ informative:
...
-flows that pass them.  This document describes a method for signaling to
-endpoints that rate limiting policies are in force and  what that rate limit is.
+flows that pass them. Access networks and their attached devices may also
+experience volatile conditions that affect the sustainable network throughput. 
+
+This document describes a method for signaling to
+endpoints that rate limiting policies are in force and what that rate limit is,
+and for non-rate limiting networks to signal throughput advice based on network
+conditions.
```
martinthomson, line 66:
```suggestion
```

See above.

draft-thoji-scone-protocol.md
```diff
@@ -57,24 +57,34 @@ informative:
...
 
 Having the network indicate what its rate limiting policy is, in a way that is
-accessible to endpoints, might allow applications to use this information when
+accessible to endpoints, allows applications to use this information when
 adapting their send rate.
 
+In addition, access networks without rate limiting policies may also provide 
+information to endpoints that can help account for connection state, such as 
+network congestion, volatile signal strength, or radio bearers with constrained
+bandwidth.
```
martinthomson, line 86: I would instead say this:

```suggestion
Network elements are not limited to communicating information
about rate limiting policies.
Network elements in access networks could provide information
to endpoints that can help account for changes in network capacity
that are not suited to congestion control feedback. This might include
reduced capacity due to overuse, equipment faults, or other transient issues; 
conversely, networks might choose to signal increased availability of capacity.
```

draft-thoji-scone-protocol.md
```diff
@@ -122,19 +132,19 @@ This protocol only works for flows that use the SCONE packet ({{packet}}).
...
 
 ## Independent of Congestion Signals
 
-Rate limit signals are not a substitute for congestion feedback.  Congestion
+throughput advice signals are not a substitute for congestion feedback.  Congestion
 signals, such as acknowledgments, provide information on loss, delay, or ECN
-markings {{?ECN=RFC3168}} that indicate the real-time condition of a network
-path.  Congestion signals might indicate a throughput that is different from the
+markings {{?ECN=RFC3168}} that suggest the real-time condition of a network
+path.  Congestion signals might suggest a throughput that is different from the
```
martinthomson, line 147: I think we need a better word than "suggest" for CC signals.

draft-thoji-scone-protocol.md
```diff
@@ -145,22 +155,22 @@ congestion controller.
...
+Endpoints can therefore be more confident in the throughput signal as an
 indication of the maximum achievable throughput than as any indication of
-expected throughput.  That throughput will only be achievable when there is no
-significant data flowing in the same scope.  In the presence of other flows,
-congestion limits are likely to determine actual throughput.
+expected throughput.  Where the throughput signal simply represents a rate
+limit, then the advised throughput will only be achievable on an uncongested network.
+Whereas a throughput signal calculated by a non rate-limiting network should be 
+based on current network and attached device signal conditions, and hence account for
+any network congestion.
```
martinthomson, line 171: I don't like this one.  The idea that this mechanism is good for congestion feedback is not something that has broad support.  That's why ECN or similar mechanisms are for.

draft-thoji-scone-protocol.md
```diff
@@ -197,12 +207,12 @@ when a network element detects a flow using more bandwidth than advertised via
...
-can be assumed to apply to the flow of packets on the same UDP address tuple for
-the duration of that flow.  Rate limiting policies often apply on the level of a
-device or subscription, but endpoints cannot assume that this is the case.  A
-separate signal can be sent for each flow.
+The time and scope over which throughput advice applies is not specified.  The
+effective throughput advice might change without being signaled.  The signaled
+advice can be assumed to apply to the flow of packets on the same UDP address 
+tuple for the duration of that flow.  For rate limiting networks, rate limiting
+policies often apply on the level of a device or subscription, but endpoints
+cannot assume that this is the case.  A separate signal can be sent for each flow.
```
martinthomson, line 215: The tentative position thus far is that we will be specifying a time period, even if we do not define a scope.  I'd rather not touch this text just now if it is changing.

draft-thoji-scone-protocol.md
```diff
@@ -331,7 +341,7 @@ QUIC endpoints can enable the use of the SCONE protocol by sending SCONE packets
 ({{apply}}) according to their policies.
 
 
-## Applying Rate Limit Signals {#apply}
+## Applying throughput advice signals {#apply}
```
martinthomson, line 344:
```suggestion
## Applying Throughput Advice Signals {#apply}
```

draft-thoji-scone-protocol.md
```diff
@@ -122,19 +132,19 @@ This protocol only works for flows that use the SCONE packet ({{packet}}).
...
 
-The rate limit signal that this protocol carries is independent of congestion
+The throughput advice signal that this protocol carries is independent of congestion
 signals, limited to a single path and UDP packet flow, unidirectional, and
 strictly advisory.
 
 ## Independent of Congestion Signals
 
-Rate limit signals are not a substitute for congestion feedback.  Congestion
+throughput advice signals are not a substitute for congestion feedback.  Congestion
```
martinthomson, line 144:
```suggestion
Throughput advice signals are not a substitute for congestion feedback.  Congestion
```

martinthomson (changes requested):
The global search and replace didn't account for capitalization properly.  The substance of other changes is also against some of the established agreements around the divide between this work and congestion control.

huitema (commented):
This PR is premature. We should have a discussion in the issue first, and when we do have consensus on the issue process a PR.

The issue discusses replacing "rate limit" by "throughput advice". We have first to agree that this is a good idea. I think it might be, but with a caveat:

1. It should be completely OK for networks that do not perform any explicit rate limit to still provide an advice, such as "this network is engineered to support at most 100Mbps per flow".

2. It is definitely not OK to expect the throughput advice to reflect instant conditions, such as "the path from New York to LA is currently congested". This is the role of congestion signals, which operate on a much shorter time scale.

draft-thoji-scone-protocol.md
```diff
@@ -145,22 +155,22 @@ congestion controller.
...
+expected throughput.  Where the throughput signal simply represents a rate
+limit, then the advised throughput will only be achievable on an uncongested network.
+Whereas a throughput signal calculated by a non rate-limiting network should be 
+based on current network and attached device signal conditions, and hence account for
+any network congestion.
+
 
-This makes the application of signals most usefully applied to a downlink flow
-in access networks, close to an endpoint. In that case, capacity is less likely
-to be split between multiple active flows.
```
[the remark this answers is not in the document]
Kevsy: > This is a true statement. What's the rationale for removing it?

Quite probably my misunderstanding - but my rationale is that  capacity will always be split between multiple active flows in a best effort access network. I'm also not clear why downlink is specified but not uplink..?

draft-thoji-scone-protocol.md
Kevsy: Thanks, that suggestion covers the motive for the PR in one paragraph :)

draft-thoji-scone-protocol.md
```diff
@@ -122,19 +132,19 @@ This protocol only works for flows that use the SCONE packet ({{packet}}).
...
 strictly advisory.
 
 ## Independent of Congestion Signals
 
-Rate limit signals are not a substitute for congestion feedback.  Congestion
+throughput advice signals are not a substitute for congestion feedback.  Congestion
 signals, such as acknowledgments, provide information on loss, delay, or ECN
-markings {{?ECN=RFC3168}} that indicate the real-time condition of a network
-path.  Congestion signals might indicate a throughput that is different from the
+markings {{?ECN=RFC3168}} that suggest the real-time condition of a network
```
Kevsy, line 144:
```suggestion
markings {{?ECN=RFC3168}} that indicate the real-time condition of a network
```

Kevsy:
```suggestion
markings {{?ECN=RFC3168}} that indicate the real-time condition of a network
```

draft-thoji-scone-protocol.md
Kevsy: Reverted to 'indicate'

Kevsy, line 145:
```suggestion
path.  Congestion signals might indicate a throughput that is different from the
```

draft-thoji-scone-protocol.md
Kevsy: Reverted to 'indicate'

draft-thoji-scone-protocol.md
Kevsy, line 169:
```suggestion
expected throughput.  That throughput will only be achievable when there is no
significant data flowing in the same scope.  In the presence of other flows,
congestion limits are likely to determine actual throughput.
```

Kevsy: Reverted (except I have kept 'throughput' in place of the original 'rate limit')

draft-thoji-scone-protocol.md
Kevsy: > +1. This PR should be limited to replacing "rate limit" by "Throughput advice".

I think it can also include Martin's text that says rate limiting is not the only use (line 83-86 in the PR), and together those would cover my motive for raising the PR

draft-thoji-scone-protocol.md
```diff
@@ -435,10 +442,10 @@ channel.
 # Security Considerations {#security}
 
 The modification of packets provides endpoints proof that a network element is
-in a position to drop datagrams and thereby enforce the indicated rate limit.
+in a position to drop datagrams and may be applying a rate limit policy.
```
martinthomson, line 445:
```suggestion
in a position to drop datagrams and could apply a rate limit policy.
```

draft-thoji-scone-protocol.md
```diff
@@ -411,15 +417,16 @@ processed.
...
 
 ## Interactions with congestion control
 
 SCONE and congestion control both provide the application with estimates
 of a path capacity. They are complementary. Congestion control algorithms
 are typically designed to quickly detect and react to congestion, i.e., to
 the "minimum" capacity of a path. SCONE informs the endpoint
-of the maximum capacity of a path.
+of the maximum capacity of a path based on network rate limit policy,
+network conditions, or a comination of the two.
```
martinthomson, line 429:
```suggestion
network conditions, or a combination of the two.
```

draft-thoji-scone-protocol.md
```diff
@@ -369,9 +375,9 @@ different ranges of bitrates. This design allows for:
 *  Graceful handling of network elements that might only recognize one version.
 
 
-## Providing Opportunities to Apply Rate Limit Signals {#extra-packets}
+## Providing Opportunities to Apply throughput advice signals {#extra-packets}
```
martinthomson, line 378:
```suggestion
## Providing Opportunities to Apply Throughput Advice Signals {#extra-packets}
```

martinthomson (commented):
With the changes, this is fine.  It's not going to be an endorsement of using the signal for real-time adaptation, like network congestion, so I'm OK with the switch to "throughput advice".

draft-thoji-scone-protocol.md
martinthomson: I'm OK with this generally now.  However, this no longer makes sense as a statement:

> The effective throughput advice might change without being signaled.

Should we drop that line?  Or is there a different statement for this?  I'm leaning toward cutting it.

kazuho: Isn't it still the case that when the advice changes while the endpoints is in quiescence, the endpoints would end up sending data without being signaled?

draft-thoji-scone-protocol.md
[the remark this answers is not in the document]
kazuho: I think the statement is suggesting that at the very last mile, it is likely that there is only one active connection at a time. I think that is true and relevant here.

kazuho (approved):
Thank you for the efforts. Have minor comments but looks good overall.

Kevsy:
Thanks @kazuho : https://github.com/ietf-wg-scone/scone/pull/31#discussion_r2090372600 explains it so I have put the text back for now, but I'll suggest a rewording in another PR to avoid bloating this PR.

draft-thoji-scone-protocol.md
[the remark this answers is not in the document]
alvestrand: The typical case that causes swearing at throughput is the adult watching TV and the teenager turning on his video game; the adult complains that the news is showing a spinner mid-stream; the teenager complains that his controller is sluggish. Multiple flows at the last mile are normal, expected, and should be engineered for.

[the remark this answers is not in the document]
Kevsy: Reverted, but I will raise another PR with suggested rewording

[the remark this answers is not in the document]
martinthomson: @alvestrand, that's what a congestion controller ultimately has to be used for.  The design here doesn't address that case, except to the extent that the sluggish game might fit between the rate limit cap and the capacity that the video uses.

Looking forward to what @Kevsy suggests here.

draft-thoji-scone-protocol.md
ihlar: > would removing 'advice' here cover it, i.e. 'The effective throughput might change without being signaled'?
To me that misses the quiescence point that @kazuho makes, since "effective throughput" sounds like something active that you measure.

Perhaps rephrasing it to something like:

“Network conditions and rate-limit policies can change in ways that make previously signaled advice obsolete, and there are no guarantees that updated advice will be sent.”

martinthomson:
@ihlar (or @Kevsy) do you think that one of you can take the task of rebasing?  GitHub is telling me that it's a bit too complicated.

ihlar (approved):

ihlar:
PR was merged manually.

PR #36, opened 2025-07-07
Reformat and tweak code

martinthomson:
The use of multiple versions makes this worse,
but I think it's worth breaking out the version change from the rest.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -355,22 +355,23 @@ rate limit; otherwise, the original values are retained, preserving the signal
 from the network element with the lower policy.
 
 The following pseudocode indicates how a network element might detect a SCONE
-packet and replace an existing rate signal.
+packet and replace an existing rate signal,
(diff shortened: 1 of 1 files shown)


kazuho:
Yeah, I think 7514f6b is fine in sense that we can assume the following:
* bit rates in SCONE2_VERSION are above those in SCONE1_VERSION, and
* within each version, bit rate signals are ordered by their bit rates.

Not that 22464ac is bad, but I wonder if network elements would actually convert signals to bit rates.

martinthomson:
There's several things in there that are probably ripe for optimization; the point of pseudocode is not to highlight what is optimal, but to describe - in an unambiguous fashion - what is correct.  For instance, the conversion from target throughput to a signal is something you can cache for the target throughput.  I didn't do that.

Given the current versions we have, there is a unique mapping from any rate to just one pair of version and signal.  That's not guaranteed more generally though.  I think that I'd prefer the more general form.

Issue #43, opened 2025-07-22
Don't require an update to every packet


martinthomson:
One equipment vendor requested that we not require network elements to update every packet.  This seems reasonable to me.

A proposed option here is that endpoints retain the lowest throughput advice received over the XX second window that we have for enforcement.  There is no sense in changing any faster than that, because it is possible that senders will have already consumed the available bandwidth at the start of any given window.

This increases the load on any endpoint because it means that you have to retain the time at which advice was last refreshed at a given level.  You might also have to retain advice at higher levels (though this is not mandatory, it would allow you to take advantage of an increase in the advice more quickly).  That seems pretty modest as far as I'm concerned.

alvestrand:
If the network marks only some SCONE probe packets, the endpoints will send more of them in order to get the signal they want. Seems like a destructive strategy to me.

martinthomson:
All of this (Nyquists theorem etc) should be addressed in the pull requests.  Note that the pull requests are all stacked.

PR #44, opened 2025-07-22
Fix the innocent protocol mangling

### LINKED ISSUE #10: Could network elements mangle innocent packets?
This is already discussed in the [TRAIN issue 32](https://github.com/martinthomson/train-protocol/issues/32). A Trone intermediary can mistake packets of some other applications for Trone packets. If it reqrite 6 bits, that will cause the packet to be eithre dropped or misinterpreted.

As explained by @ihlar during the IETF meeting in Bangkok, we need to distinguish:

1. random collision: the end to end application issues some random first bytes that happen to mimic Trone. It will happen at low frequency.
2. systematic collision: the end to end collision had negotiated some header that happens to Mimic Trone. All packets with that header will be mangled.

The proposed solution is for intermediaries to keep flow state and stop marking if too many packets of a flow match Trone. We need to develop that.

martinthomson:
This depends on an answer to #43 being as I suggested there. I won't make the pull request dependent on that though.

Closes #10.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -658,6 +658,7 @@ to reduce the reaction time from senders.
 In addition, some heuristics might be used
 to detect SCONE-compatible QUIC flows.
 This includes identification of a QUIC handshake on the flow,
+the presence of indications ({{indication}}),
 or other heuristics.
 If these heuristics indicate a non-QUIC flow,
 the safest option is
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -647,10 +647,15 @@ because that could mean that the protocol is
 effectively unable to operate end-to-end.
 
 To that end, network elements MUST ensure that
-they do not update the content of datagrams
-any more than a small number of times over each enforcement period.
+they do not update the content of every datagram.
+Limiting the rate of updates
+to a small number of times over each enforcement period
+is sufficient.
+Network elements might update more often
+in response to a change in throughput advice,
+to reduce the reaction time from senders.
 
-Some heuristics might be used
+In addition, some heuristics might be used
 to detect SCONE-compatible QUIC flows.
 This includes identification of a QUIC handshake on the flow,
 watching to ensure that the ratio of the two SCONE versions
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -648,7 +648,7 @@ effectively unable to operate end-to-end.
 
 To that end, network elements MUST only update the content of datagrams
 on a given address tuple
-more than a few times each enforcement period.
+a few times each monitoring period.
 Network elements MAY update more often
 immediately after a change in their throughput advice,
 to reduce the reaction time from senders.
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -658,8 +658,7 @@ to reduce the reaction time from senders.
 In addition, some heuristics might be used
 to detect SCONE-compatible QUIC flows.
 This includes identification of a QUIC handshake on the flow,
-watching to ensure that the ratio of the two SCONE versions
-is close to 1 over time, or other heuristics.
+or other heuristics.
 If these heuristics indicate a non-QUIC flow,
 the safest option is
 for network elements to disable updating of datagrams.
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -646,7 +646,7 @@ The most serious damage occurs when every datagram is modified,
 because that could mean that the protocol is
 effectively unable to operate end-to-end.
 
-To that end, network elements can ensure that
+To that end, network elements MUST ensure that
 they do not update the content of datagrams
 any more than a small number of times over each enforcement period.
 
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -646,13 +646,11 @@ The most serious damage occurs when every datagram is modified,
 because that could mean that the protocol is
 effectively unable to operate end-to-end.
 
-To that end, network elements MUST ensure that
-they do not update the content of every datagram.
-Limiting the rate of updates
-to a small number of times over each enforcement period
-is sufficient.
-Network elements might update more often
-in response to a change in throughput advice,
+To that end, network elements MUST only update the content of datagrams
+on a given address tuple
+more than a few times each enforcement period.
+Network elements MAY update more often
+immediately after a change in their throughput advice,
 to reduce the reaction time from senders.
 
 In addition, some heuristics might be used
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-protocol.md
```diff
@@ -528,6 +528,34 @@ a processing error at network elements. For example, they might pick connection
...
+is sufficient.
+Network elements might update more often
+in response to a change in throughput advice,
+to reduce the reaction time from senders.
+
+In addition, some heuristics might be used
+to detect SCONE-compatible QUIC flows.
+This includes identification of a QUIC handshake on the flow,
+watching to ensure that the ratio of the two SCONE versions
+is close to 1 over time, or other heuristics.
```
martinthomson, line 554:
```suggestion
or other heuristics.
```

huitema (commented):
I am concerned that this mitigation requires the network elements to keep state by 4-tuples . That could be onerous, and could also open the door for denial of service attacks.

The other possible mitigation is to change the scone format to add entropy and reduce the risk of collision. I think that we discussed a solution like that before. For example, add a payload to the scone packet, with two numbers: a counter, and a proof of that counter such as XORing it with the version number or some other magic number. At that point the risk of repeated collision becomes really low. 

Of course, we have a tradeoff here. On one hand, require operators of network elements to keep some state, on the other hand, require some more complicated processing. Both fall in the category of "you must do that but I know you wont", with lazy operators either forgetting about checks and processing every packet that they recognize as SCONE, or just rolling the dice for every packet recognized as SCONE to decide whether to mark it.

And of course, the "rolling the dice" strategy implies a heavy probability of SCONE packets not marked, which will induce end points to send way more SCONE packets than necessary. Which in turn will stress the network elements.

draft-ietf-scone-protocol.md
```diff
@@ -528,6 +528,34 @@ a processing error at network elements. For example, they might pick connection
...
+is sufficient.
+Network elements might update more often
+in response to a change in throughput advice,
+to reduce the reaction time from senders.
+
+In addition, some heuristics might be used
+to detect SCONE-compatible QUIC flows.
+This includes identification of a QUIC handshake on the flow,
+the presence of indications ({{indication}}),
+or other heuristics.
```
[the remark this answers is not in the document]
martinthomson: Yeah, I expect #49 to merge first and then this is fine.

martinthomson:
So the only real requirement here is not to alter every packet.  That, of course, depends on #46, but I think that's necessary for other reasons (robustness primarily).  The heuristics only apply where there is per-flow state, not everywhere.  Is there a caveat that would make that part more acceptable @huitema?

huitema:
Assume a lazy implementer that does not want to keep state, so simply picks a random number with some probability to decide whether marking the SCONE packet. Now, suppose a systematic collision. If the probability of marking is `x`, then the fraction of mangled packets is also `x`. What is the acceptable rate of mangled packets? I think lower than 1% is probably OK, TCP is more or less designed to handle that. More than that will be severe. 10% means serious performance degradation. 50% means fully broken.

At the same time, picking a random rate of 1% means a terrible SCONE service. 50% would probably work, leading to endpoints sending twice as many SCONE packets as required. 10% would lead to 10 times as many packets, and that seems bad.

My preference would be to not mention a fraction, but rather mandate a delay. We have seen that if the signal is good for 60 seconds, endpoints should probably sample it every 30 seconds. That means repetitions at shorter interval probably indicate brokenness. If I had to implement that, I would probably use a Bloom filter to check whether the 4-tuple has been seen recently, and in that case leave the packet alone. And I would just zero the filter at short intervals, maybe 5 seconds, maybe 10. That seems something even lazy folks can implement, especially if we twist their arm.

My recommendation would be to say something like MUST NOT process a "SCONE" packet if the 4-tuple has been seen in the past N seconds. Pick a number for N -- 2, 5, 10 probably all work. Then even the lazy folks can think about a Bloom filter implementation, or something like that.

martinthomson:
> My preference would be to not mention a fraction, but rather mandate a delay. We have seen that if the signal is good for 60 seconds, endpoints should probably sample it every 30 seconds.

Isn't that what this does?

> Limiting the rate of updates to a small number of times over each enforcement period is sufficient.

Unless you feel like "small number" needs to be more concrete (2-10 seems reasonable, as you say).

draft-ietf-scone-protocol.md
```diff
@@ -528,6 +528,34 @@ a processing error at network elements. For example, they might pick connection
...
+effectively unable to operate end-to-end.
+
+To that end, network elements MUST ensure that
+they do not update the content of every datagram.
+Limiting the rate of updates
+to a small number of times over each enforcement period
+is sufficient.
+Network elements might update more often
+in response to a change in throughput advice,
+to reduce the reaction time from senders.
```
martinthomson, line 548:
```suggestion
To that end, network elements MUST only update the content of datagrams
on a given address tuple
more than a few times each enforcement period.
Network elements MAY update more often
immediately after a change in their throughput advice,
to reduce the reaction time from senders.
```
Tweaking @huitema's suggestion.

draft-ietf-scone-protocol.md
```diff
@@ -528,6 +528,32 @@ a processing error at network elements. For example, they might pick connection
...
+because that could mean that the protocol is
+effectively unable to operate end-to-end.
+
+To that end, network elements MUST only update the content of datagrams
+on a given address tuple
+more than a few times each enforcement period.
+Network elements MAY update more often
+immediately after a change in their throughput advice,
+to reduce the reaction time from senders.
+
```
huitema, line 547: OK, we are down to editorial. I suggested "SHOULD NOT ... more than a few times". You propose "MUST only ... more than a few times." I still prefer my original proposal for two reasons:

- "MUST only ... more" is a weird construct. I think this should be either "MUST NOT ... more" or "SHOULD NOT .. more".
- I originally thought of writing "MUST NOT", but in general if we have a combination of mandate and exception we write "SHOULD NOT \<do something>" followed by "MAY \<by exception do something else>", so I would rather fit the established pattern.

draft-ietf-scone-protocol.md
```diff
@@ -528,6 +528,32 @@ a processing error at network elements. For example, they might pick connection
...
+In addition, some heuristics might be used
+to detect SCONE-compatible QUIC flows.
+This includes identification of a QUIC handshake on the flow,
```
kazuho, line 550: Considering that SCONE is meant to work with any version of QUIC, I'm not sure if we would want to talk about identifying handshakes, as that is a concept specific to version 1 of QUIC.

martinthomson: I think that this is fine in the context of applying heuristics.

draft-ietf-scone-protocol.md
martinthomson: Done.  Thanks.

martinduke (approved):

ihlar:
The risk of protocol mangling could be an argument for creating a maximum sending frequency of SCONE packets. If a network element counts more than N SCONE packets during a monitoring period it could just stop updating datagrams on that 4-tuple. 
The current text already mitigates the potential harm quite well, but this would reduce the potential harm even further.

martinthomson:
@ihlar, maybe open an issue.  I'm not sure that this is necessary (consistent SCONE detection is unlikely and not every packet is updated), but setting a floor on the addition of SCONE packets and also the updating of the rate signal would be more robust.  Whether we need that robustness is really the key question.

ihlar (approved):

PR #46, opened 2025-07-23
Add a minimum rate algorithm

martinthomson:
This approach ensures that we don't have blackout periods (an issue raised in the meeting).  But it makes the whole rate limit signaling more robust.  It also allows network elements the option of skipping updates.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -390,50 +390,21 @@ applications MUST apply the lowest throughput advice
 received during any monitoring period; see {{time}}.
 
 Senders can ensure that their use of network capacity
-remains within throughput advice
-using a lowest rate algorithm as follows.
+remains within throughput advice.
 
-To operate this algorithm,
-the minimal state that an application --
-which is often, but not always a sender --
-maintains is
-* the current rate limit,
-* any state necessary to monitor throughput
-  (that is, throughput usage state, such as the state in {{sliding-window}}), and
-* the time at which that rate limit was last updated.
-
-When updated advice is received:
-
-* If the signaled rate is the same as the current rate,
-  set the last update time to the current time.
-
-* If the signaled rate is higher than the current value,
-  then:
-
-  * If the time since the last update is greater than the monitoring period,
-    the current rate limit is increased to the received value
-    and the last update time is set to the current time.
-
-  * Otherwise, the throughput advice is not applied.
-    An implementation can optionally remember throughput advice,
-    so that it might increase send rates more quickly,
-    but this is not necessary
-    as long as there are frequent opportunities
-    to receive updated throughput signals.
-
-* If the signaled rate is lower than the current value,
-  set the current rate limit to match the advice,
-  optionally discard all rate monitoring state,
-  and set the last update time to the current time.
-
-Applications MAY discard throughput usage state when reducing rates.
+Applications MAY discard throughput usage state
+when they receive throughput advice that indicates a reduced rate.
 Otherwise, data that was sent at a higher rate
-might force the available rate to zero; see also {{policing}}.
-Even though loss of information about recent send rate
-might result in temporarily exceeding the rates indicated by throughput advice,
-the relatively long duration of the monitoring period
+might force the available rate to zero
+for the remainder of the monitoring period; see also {{monitoring}}.
+
+The relatively long duration of the monitoring period
 means that this is preferable to disabling sending completely.
-Applications retain throughput usage state when increasing state.
+The cost is that loss of information about recent send rate
+might result in temporarily exceeding the rates indicated by throughput advice.
+In comparison,
+applications retain throughput usage state
+when increasing state.
 
 This approach ensures that network elements
 are able to reduce the frequency with which they send updated signals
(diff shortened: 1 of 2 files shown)


draft-ietf-scone-protocol.md
```diff
@@ -425,7 +489,13 @@ if is_long and (packet_version == SCONE1_VERSION or
       packet[1..5] = htonl(target_version)
 ~~~
 
-## Flows That Exceed Throughput Advicea
+A network element needs to ensure that it sends updated rate signals
+with no more than a monitoring period ({{time}}) between each.
+Because this depends on the availability of SCONE packets,
+network elements might need to update at a much higher frequency.
+
```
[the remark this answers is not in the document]
martinthomson: I think that this randomization can come later.

wesley-eddy:
I think this is missing the mark a bit or making things overly complex.

The problem it's aiming to solve might be better discussed in the WG a bit more to make sure the understanding is aligned with vendors and CSPs that would be implementing and deploying, to make sure it's aligning with needs.

The notion of monitoring periods, sampling, nyquist, etc. are not needed.  The max throughput guidance is not like an analog signal that we're trying to sample and reproduce with some fidelity.  For most envisioned current cases, it's a fixed number that will not change for the duration of a flow.  It comes from the mobile user's subscriber plan (e.g. the plan says video at 720p is supported, so the SCONE signaling always indicates a corresponding fixed rate).

The text here about senders and receivers keeping state is really adding too much complexity for the typical case where a client gets SCONE advice of N Mbps, and then is able to use that knowledge in its ABR algorithm as a top limit subsequently for video chunks requested in the flow.  Allowing it to refresh or receive updated guidance does not have to be this complex IMHO.

huitema:
I think we need to balance a few tensions there:

1. Time limited advice. The conditions will change. They may change randomly. The application will usually get some advice if conditions change for the worse, but probably not if they change for the better. They could easily trigger a SCONE probe in case of worsening, but they do have to resort to periodic or quasi periodic probing to detect easing. The PR helps here with a number, 67. If the advice is supposed good for 67 seconds, the client has no reason to probe more than every 20 to 30 seconds.

2. Load of network intermediaries. Ideally, intermediaries should be able to process all incoming packets at line rate, but that may not be realistic. Stateless solutions such as only marking a random fraction of packets wont work well: if the application sends 3 probes in 67seconds, any significant non marking rate incurs the risk of none of these 3 packets being processed. Thus, we need some state per flow.

3. Difficulty of keeping state in network intermediaries. The number of flow could be very large, but if applications only send a stone packet every 20-30 seconds and send video at 720p, that's one mark per 2000 to 3000 packets. This can be used as a design criteria -- if the intermediary monitors a link carrying 1M pps, it should be able to handle 1K SCONE per second. That can be used for example to dimension a Bloom filter. The network element can use a Bloom filter that is zeroed every 10 seconds, check whether the flow is already in the filter, not process the packet if it is there, process the SCONE packet and add a mark in the filter if it is not.

Bottom line, I think the advice is correct.

martinthomson:
I've updated this pull request with an implementation (and tests) for the algorithm.  It really is very simple to implement:

```python
  def set_advice(self, t, advice):
    """Update the throughput advice based on a received signal."""

    if advice < self.advice:
      # Flush state if the advice reduces.
      self.state = [0] * self.count
      self.total = 0
    elif self.advice_updated + self.interval * self.count > t:
      # Ignore increases too soon after an update.
      # This could be smarter about increases, but this is safe.
      return
    self.advice = advice
    self.advice_updated = t
```

kazuho:
@martinthomson I'm not sure if the snippet (or the script included in the PR) is correct.

Consider the case following case:
1. At t=0, an endpoint receives an advice of 1Mbps.
2. The endpoint notifies the peer and immediately receives from peer `1Mb * duration` bits of data.
3. At t=1, the endpoint receives an advice of 500Kbps.

At this moment, how much additional budget does the endpoint have? I would argue that the budget is zero. However, based on the proposed snippet, I think the endpoints would assume it can receive additionally `500Kb * duration` bits of data?

draft-ietf-scone-protocol.md
```diff
@@ -378,6 +379,69 @@ the throughput advice that is received on datagrams with one marking
...
+using a lowest rate algorithm as follows.
+
+To operate this algorithm,
+the minimal state that a sender maintains is
+* the current rate limit,
+* any state necessary to monitor throughput
+  (that is, throughput usage state, such as the state in {{sliding-window}}), and
+* the time at which that rate limit was last updated.
+
+When a SCONE packet containing updated advice is received:
```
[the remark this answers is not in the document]
martinthomson: Too complex?  It's essential to understand what the expectations are.  Having an algorithm like this makes it precise and the expectations clear.

Your point about what receives the signal is well-taken though.  I've modified this to not focus on the specific role.

martinthomson:
Thanks for the feedback.  I've done as Alan suggests and added a bunch of advice on how to send SCONE packets (because it fit).  Those who were unhappy with this should take another look.

draft-ietf-scone-protocol.md
```diff
@@ -506,8 +543,80 @@ if is_long and (packet_version == SCONE1_VERSION or
...
+A network element needs to ensure that it sends updated rate signals
+with no more than a monitoring period ({{time}}) between each.
+Because this depends on the availability of SCONE packets
+and packet loss can cause signals to be missed,
+network elements might need to update more often.
+
+Senders that send a SCONE packet
+or network elements that update SCONE packets
+every 20&ndash;30 seconds is likely sufficient to ensure that throughput advice is not lost.
+Senders can avoid synchronization of SCONE packets,
```
afrind, line 555: A few more words explaining the kind of synchronization to be avoided might add clarity - I had a to read this a couple times before I realized what you meant.

draft-ietf-scone-protocol.md
```diff
@@ -506,8 +543,80 @@ if is_long and (packet_version == SCONE1_VERSION or
...
+which might cause network elements to miss updates,
```
afrind, line 556: I'm not sure why synchronization might cause missed updates, but maybe it doesn't matter.

draft-ietf-scone-protocol.md
```diff
@@ -506,8 +543,80 @@ if is_long and (packet_version == SCONE1_VERSION or
...
+
+To operate this algorithm,
+the minimal state a network element maintains is:
+* the current rate limit,
+* any state necessary to monitor throughput
+  (that is, throughput usage state, such as the state in {{sliding-window}}),
+* a timer used for rate increases, and
+* the time at which that rate limit was last updated,
+
+When advice is updated,
```
afrind, line 576: Maybe "set or updated" to capture the first time advice is given?

draft-ietf-scone-protocol.md
```diff
@@ -550,8 +659,51 @@ discards a SCONE packet without also successfully processing another packet
 from the same datagram SHOULD ignore any throughput advice signal. Such a datagram
 might be entirely spoofed.
 
+Once negotiated, endpoints that seek to receive throughput advice on a flow
+MUST send a SCONE packet at least twice each monitoring period; see {{time}}.
+
+Sending SCONE packets more often might be necessary to:
```
afrind, line 665: I'm trying to understand the consequence of not conforming to the above MUST (2x/period).  I assume it's not "or the NE drops the flow".

I read this bit and the section that follow as why a sender should send than 2x/period, but maybe its just line 665 that is misleading me?

draft-ietf-scone-protocol.md
```diff
@@ -550,8 +659,51 @@ discards a SCONE packet without also successfully processing another packet
...
+A real-time conferencing application might send more often.
+In either case, the length of the monitoring period ({{time}})
+limits how fast any application can react.
+
+Though sending SCONE packets more than once each round trip time
+might help reduce exposure to packet loss,
+it is better to spread updates over time
+rather than to send multiple SCONE packets in less frequent bursts.
+
+The main cost associated with sending SCONE packets
```
afrind, line 700: Is there any compute cost on the network element that we should mention?

afrind (commented):
Thanks for the update, this seems much better to me.

britram:
from the 26 August interim: 

- please review language to make sure it's clear this is guidance, client behavior doesn't need to be standardized for interop
- consider how this operates with occasional drops / missing advice
- wait until EOD AOE then merge it.

We'll spend some time testing this in Montreal

draft-ietf-scone-protocol.md
martinthomson: I am going to assume that the cost is modest because the network element can (now) ignore most SCONE packets.  Any they do update will require effort, but that's effort that they balance against the benefits they deliver (better user experience, better adherence to network policy, etc...).

PR #48, opened 2025-07-23
Forbid insertion of SCONE packets

### LINKED ISSUE #21: Network inserted TRONE packets
The topic of whether network elements should be able to generate TRONE packets and insert them into expanded datagrams was discussed at IETF 122. Also discussed in #6.
Ideally this should not be needed, given that there is proper guidance on when to send TRONE packets and making sure that updating the rate limit field of the TRONE packets is sufficiently cheap. The risk of expanding datagrams in the network is that it can result in path MTU issues with lost packets as a consequence. 

There are a few paths we can take on this topic:

1. Say nothing about it.
2. Mention the possibility and give guidance on potential MTU issues that can arise. 
3. Explicitly forbid network inserted TRONE packets and design protocol mechanisms to enforce it.

martinthomson:
This is one potential approach to #21.  However, it does constrain things more than I think is necessary.  I tend to think that staying quiet is best, because it gives network elements the option and endpoints the choice of ignoring those inserted SCONE packets.  I wouldn't want to *encourage* that outcome though.

Closes #21.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,5 @@ Gemfile.lock
 archive.json
 package-lock.json
 report.xml
+ta.py.excerpt
 !requirements.txt
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 /.refcache
 /.venv/
 /.vscode/
+/__pycache__/
 /lib
 /node_modules/
 /versioned/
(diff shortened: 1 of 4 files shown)


britram:
guidance from August 26th interim: ship it, close 21.

martinthomson:
Closing in favour of #62.

PR #50, opened 2025-07-23
Define the transport parameter

### LINKED ISSUE #41: value of the scone_supported transport parameter
At the moment, the draft is silent regarding what the value of the scone_supported parameter should be.

Assuming that this parameter is boolean, I believe the value should be zero-length, like the disable_active_migration TP defined in RFC 9000.

martinthomson:
Give it a value and a length (0).

Closes #41.


ihlar (approved):

huitema (approved):
Yes. Empty TP is the simplest design.

mjoras (approved):

Issue #45, opened 2025-07-23
create a 7-bit rate signal field, by allocating two QUIC versions that differ only with MSB


kazuho:
SCONE uses a 7-bit value (128 code points) for representing the rate signal.

At the moment, the most significant bit of the rate signal affects the selection of the QUIC version, and the rest of the bit is stored in the first byte of the datagram.

This construction is sub-optimal and is a pain in terms of both specification and implementation.

I propose to allocate two QUIC versions that are identical for all bits expect the most significant bit.

```
SCONE Packet {
  Header Form (1) = 1,
  Reserved (1),
  Rate Signal (7),
  Common Bits of SCONE Versions (31) = 0xSCONE1 & 0x7fffffff,
  Destination Connection ID Length (8),
  Destination Connection ID (0..2040),
  Source Connection ID Length (8),
  Source Connection ID (0..2040),
}
```

With this change, detecting SCONE packets becomes as simple as `(first64bit & 0xc07fffffffffffffff) == 0xSCONE1 & 0x7fffffff)`, and the rate signal becomes a 7-bit value that can be directly processed (without looking up different base values based on the QUIC version).

PR #54, opened 2025-07-30
MSB of version is LSB of the rate signal

kazuho:
This PR is against #51, uses MSB of the version field to carry the LSB of the rate signal, so that the bit 2 to bit 8 of the first 64-bit field carries a rate signal, which is calculated by using a logarithmic function without bit swaps or any other complexity.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -344,7 +344,7 @@ and the corresponding throughput advice for each.
 | 112 Gbps   | 60          | 0xef7dc0fd |
 | 199.5 Gbps | 63          | 0x6f7dc0fd |
 | No limit   | 63          | 0xef7dc0fd |
-{: #ex-rates title="Examples of signals and corresponding rates"}
+{: #ex-rates title="Examples of SCONE signals and corresponding rates"}
 
 
 ## Endpoint Processing of SCONE Packets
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
I don't mind either way on this one, though I will note that this is in some ways *simpler* than before (so consider this an approval).

To validate this, the code starts looking pretty lean:

```rust
        // Setup involves math
        let v = (((target_rate / 100_000.0).log10() * 20.0) as u64).clamp(0, 63);
        let target_value = (0x106f << 48) | (v << 61);

        // updates
        let x = u64::from_be_bytes(packet.first_chunk().unwrap().to_owned());
        if (x & 0x807f_ffff_ff00_0000) == 0x106f_7dc0_fd00_0000
            && target_value < (x & 0xbfff_0000_0000_0000)
        {
            packet[0] = (packet[0] & 0x40) | (target_value >> 56) as u8;
            packet[1] = (target_value >> 48) as u8;
        }
```

kazuho:
@ihlar WDYT?

ihlar (approved):
I think this approach is fine. Implemented it and it definitely makes the implementation even more lean and efficient. 

I think we need to consider whether we need to explicitly talk about two versions now. In my implementation I only need to care about the lower 31 bits (0x6f7dc0fd). Since the actual rate signal 7 contiguous bits, I think it would make since to define the Rate Signal Field to be that and the version bits as a 31 bit field in the SCONE packet Format. We could then also have a nicer table with examples between 0 and 127.  

I see that's exactly what @kazuho proposes in #45. I'd prefer that approach, but we can merge this first and continue that discussion separately.

martinthomson:
OK, I'm going to do that.  Once I resolve all these conflicts.

kazuho:
Thank you!

PR #59, opened 2025-08-21
Update format description

ihlar:
Editorial proposal where focus is on the 7-bit rate signal field as opposed to the 6-bits + version focus. 
We can discuss whether we want to do anything with the pseudo code as well.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -279,17 +279,15 @@ The most significant bit (0x80) of the packet indicates that this is a QUIC long
 header packet.  The next bit (0x40) is reserved and can be set according to
 {{!QUIC-BIT=RFC9287}}.
 
-The Rate Signal field uses 7 bits spanning the low 6 bits (0x3f) of the first
-byte and the most significant bit of the version field. This creates a
-contiguous 7-bit field with values from 0 to 127. Values for this field are
-described in {{rate-signal}}.
+The Rate Signal High Bits field consists of the low six bits (0x3f) of the
+first byte. Together with the most significant bit of the Version field,
+this forms the 7-bit Rate Signal. Values for the Rate Signal are described in
+{{rate-signal}}.
 
-The remaining 31 bits of the version field contain the common version bits
-0x6f7dc0fd. The complete version field is either 0x6f7dc0fd (when the
-least significant bit of the Rate Signal is 0) or 0xef7dc0fd (when the least
-significant bit of the Rate Signal is 1). This design facilitates detection
-and modification of SCONE packets while maintaining compliance with the QUIC
-invariants.
+The Version field contains either 0x6f7dc0fd or 0xef7dc0fd. The only difference
+between these two values is the most significant bit, which also contributes to
+the Rate Signal. All other bits are identical, which facilitates detection and
+modification of SCONE packets.
 
 This packet includes a Destination Connection ID field that is set to the same
 value as other packets in the same datagram; see {{Section 12.2 of QUIC}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-protocol.md
```diff
@@ -260,8 +260,8 @@ from {{Section 4 of INVARIANTS}}.
 SCONE Packet {
   Header Form (1) = 1,
   Reserved (1),
-  Rate Signal (6),
-  Version (32) = 0x6f7dc0fd or 0xef7dc0fd,
+  Rate Signal (7),
+  Common version bits (31) = 0x6f7dc0fd,
```
kazuho, line 264: +1 to @martinthomson that the definition of these fields should better be consistent with that of RFC 8999 and 9000. At the same time, +1 to @ihlar to describe the signal as a 7-bit value, rather than a combination of a 6-bit field and the two QUIC versions.

Therefore, maybe change the field definitions to something like below, and explain that the Rate Signal is conveyed as a 7-bit value spanning across the two consecutive fields? Then, in the rest of the document, we can just talk about the "Rate Signal."

```suggestion
  Lower 6 Bits (6),
  Version (32) = 0x6f7dc0fd or 0xef7dc0fd,
```

martinthomson:
```suggestion
  Rate Signal High Bits (6),
  Version (32) = 0x6f7dc0fd or 0xef7dc0fd,
```

Then we can say a Rate Signal comprises "Rate Signal High Bits" and an LSB which is the MSB of the "Version" field.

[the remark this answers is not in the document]
huitema: +1
Scone is defined to be compatible with QUIC invariants, and the invariants describe the version as 32 bits. Let's keep it that way. Then maybe redefine _Rate Signal (6)_ as _Rate Signal_MSB (6)_, and add text on how to extract the _Rate Signal (7)_ from the MSB and the top bit of version.

huitema: Yes. That's exactly what I suggested in my previous comment.

draft-ietf-scone-protocol.md
```diff
@@ -279,12 +279,17 @@ The most significant bit (0x80) of the packet indicates that this is a QUIC long
 header packet.  The next bit (0x40) is reserved and can be set according to
 {{!QUIC-BIT=RFC9287}}.
 
-The low 6 bits (0x3f) of the first byte contain the Rate Signal field. Values
-for this field are described in {{rate-signal}}.
+The Rate Signal field uses 7 bits spanning the low 6 bits (0x3f) of the first
+byte and the most significant bit of the version field. This creates a
+contiguous 7-bit field with values from 0 to 127. Values for this field are
+described in {{rate-signal}}.
```
huitema, line 285: Can we tie that to the proposed definition of the field as _Rate Signal High Bits (6)_ ?

draft-ietf-scone-protocol.md
```diff
@@ -300,50 +305,45 @@ header cannot precede any other packets.
 
 ## Rate Signals {#rate-signal}
 
-The Rate Signal field in SCONE uses the low 6 bits (0x3f) of the first byte.
-This field is encoded as a logarithmically spaced distribution over a range
-defined by the SCONE protocol version.
+The Rate Signal field in SCONE uses 7 bits: the low 6 bits (0x3f) of the first
+byte combined with the most significant bit (0x80) of the second byte.
+This creates a contiguous 7-bit field with values from 0 to 127.
```
huitema, line 311: Same comment. If we end up calling the 6 bits Rate _Signal High Bits (6)_ in the diagrams, lets reuse the name here.

huitema (commented):
The 7 bit field is a great idea, but it is important that we remain compatible with QUIC invariants.

draft-ietf-scone-protocol.md
ihlar: Fixed in latest commit.

huitema (approved):
Looks good after the recent changes.

PR #60, opened 2025-08-25
use 7-bit rate signal in pseudo code

ihlar:
Given the current design, it makes sense to compare the rate signals directly. 
This is not the most optimal code, but trying to prioritize readability.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -479,7 +479,7 @@ with an encoded throughput advice (`target_signal`).
 is_long = packet[0] & 0x80 == 0x80
 packet_version = ntohl(packet[1..5])
 if is_long and packet_version &0x7fffffff == SCONE1_VERSION:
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-protocol.md
```diff
@@ -473,22 +473,16 @@ from the network element with the lower policy.
 
 The following pseudocode indicates how a network element might detect a SCONE
 packet and replace an existing rate signal,
-given throughput advice (`target_throughput`).
+with an encoded throughput advice (`target_signal`).
```
huitema, line 477: Doesn't read very well. Suggestion: "... replace an existing rate signal (`packet_signal`) with a new rate signal encoding the throughput advice of this network element (`target_signal`)."

huitema (commented):
On the right track, but the sentence introducing the pseudo-code could be clarified.

huitema (approved):

PR #61, opened 2025-08-26
Monitoring and time changes

martinthomson:
This is a big set of changes, discussed in #46 and #37, which started to get a little messy.  This includes:

1. Defining a monitoring period is 67 seconds.  Network elements can monitor usage for adherence to the advice they give over this period or longer.  Endpoints can consider this to be the upper bound on the window they spread their usage over.
2. Describing a sample process for monitoring adherence.  As an example only.
3. Allowing network elements to miss updates, but recommending that they update at least twice each monitoring period; or, more specifically, every 20-30s.
4. Recommending that endpoints follow the lowest throughput advice they have received in the last 67s.
5. Allowing endpoints to discard state when the advice reduces.
6. Advising endpoints on how often to include SCONE packets and why (to avoid missing advice and to reduce latency when advice changes).

This is what we discussed during the meeting.  The direction we agreed to broadly was to give this a spin and try to improve it.  There are some small tweaks we discussed that can be pursued separately, but this batch was pretty substantial already.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -217,9 +217,11 @@ when a network element detects a flow using more bandwidth than advertised via
 SCONE, it might switch to applying its policies for non-SCONE flows, using
 congestion control signals.
 
-The signaled advice can be assumed to apply
-to the flow of packets on the same UDP address tuple
-until the advice is updated or the flow ends.
+The signaled advice applies to the flow of packets
+on the same UDP address tuple for the duration of
+the current monitoring period, unless it is updated
+earlier or the flow ends; see {{time}} for details on
+the monitoring period.
 Rate limiting policies often apply on the level of a device or subscription, but endpoints
 cannot assume that this is the case.  A separate signal can be sent for each flow.
 


martinduke (approved):
LGTM

draft-ietf-scone-protocol.md
```diff
@@ -494,6 +566,108 @@ if is_long and (packet_version == SCONE1_VERSION or
...
+
+In the event that a flow exceeds these limits,
+a network element could immediately start
+enforcing adherence to the advice as though it were a hard limit.
+However, this risks creating a situation where communication ceases entirely
+for a significant period of time;
+that is, up to the period defined in {{time}}.
+
+A better approach is to disregard any data that was transmitted
+before engaging any hard limits to throughput.
```
martinthomson, line 668:
```suggestion
A better approach is to exclude any data transmitted
before sending reduced throughput advice
from any monitoring that uses the reduced rate.
```

draft-ietf-scone-protocol.md
```diff
@@ -494,6 +566,108 @@ if is_long and (packet_version == SCONE1_VERSION or
...
+the network element waits until
+it receives the next SCONE packet on affected flows.
+It then updates the throughput advice in that packet ({{apply}})
+and updates its monitoring state as follows:
+
+* If the signaled rate is the same as the current rate,
+  set the last update time to the current time.
+
+* If the signaled rate is higher than the current value,
+  then:
```
kazuho, line 620: PS. As a high order point, I wonder the proposed algorithm is incorrect in sense that it retains old limit iff the rate limit is increased.

IMO, the correct approach _for network elements_ is to retain old limits when the rate limit is reduced, but apply the new rate limits immediately when they are increased.

This is in contrast to how _endpoints_ should behave; i.e., is retain old limits when the rate limit is increased, but apply the new limits ASAP when they are decreased.

martinthomson: It retains the old rate if the new rate is the same or higher (not "iff").

How could you retain the old limit if the new one reduced?

Let's say that you start at 1Mbps and send at that rate for an extended time.  At the point the number drops to 200Kbps, you have just spent the last however long sending at a higher rate.  Let's assume that the send rate is approximately constant.  In that case, because the rate dropped to 1/5th of the previous, if any period you measure includes the period of period/5 seconds prior to the change in advice, you cannot send anything at all 4/5ths of a monitoring period.

I think that I'm going to need you to more clearly explain exactly how you imagine endpoints and network elements behave here, because I'm clearly not understanding your point.  I don't know how to implement:

> IMO, the correct approach for network elements is to retain old limits when the rate limit is reduced, but apply the new rate limits immediately when they are increased.

martinthomson:
@huitema
> Maybe we should say that the maximum pacing rate should be no more that Y times the advice, and discuss whether Y should be 1, 2, 8, 67 or something more.

My understanding was that such bursts were in fact desirable in many networks, modulo actual achievable rates, congestion control, buffers, etc...  Because it does in fact allow endpoints to go to sleep.

Moreover, real congestion controllers don't allow that to happen.  They just can't be quiet for extended periods (like 30 seconds) and immediately switch to sending at line rate.  So you have a smoothing function present anyway to protect the network.  Even the most aggressive resumption 

Those combined mean that I'm not concerned about this leading to a real problem.  I don't think it's crazy to put a secondary cap on send rate that is some multiple of the baseline rate, but I expect that the number could be quiet high.

draft-ietf-scone-protocol.md
```diff
@@ -494,6 +554,108 @@ if is_long and (packet_version == SCONE1_VERSION or
...
 
+A network element needs to ensure that it sends updated rate signals
+with no more than a monitoring period ({{time}}) between each update.
+Because this depends on the availability of SCONE packets
+and packet loss can cause signals to be missed,
+network elements might need to update more often.
+
+Senders that send a SCONE packet
+or network elements that update SCONE packets
+every 20&ndash;30 seconds is likely sufficient to ensure that throughput advice is not lost.
```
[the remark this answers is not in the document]
kazuho: I see, thank you for the clarification.

To me it seems that "add" in "senders can _add_ a small random delay" was the source of my confusion. Could we change that to "include"?

martinthomson:
Given the question about bursting, I think that it's worth taking that up separately.  I'll open an issue for that.

draft-ietf-scone-protocol.md
kazuho: > I think that I'm going to need you to more clearly explain exactly how you imagine endpoints and network elements behave here, because I'm clearly not understanding your point.

Let's reuse the previous example.

1. [t=0] NE indicates 1Mbps
2. [t=0] endpoints start sending at 1Mbps, which is expected to continue for 50 seconds
3. [t=20] NE indicates 100Kbps

Consider this endpoint a real-time video application. By receiving a bit rate advice via a SCONE packet at t=0, it is assuming that it can send at 1Mbps. The application updates its send rate only once every 50 seconds, because 50 is smaller than 67.

As the application updates its send rate every 50 seconds, it fails to adopt to the bit rate reduction at t=20. The logic being proposed in this pull request would call out excess usage at t=26.7; i.e., 6.7 seconds after the 100Kbps signal is sent.

Note that the grace period in this example is 6.7 seconds, because the rate limit was reduced to a 1/10th of the original. As the ratio of reduction is sharper (it could theoretically be as large as 100K/199.5G), the "grace period" becomes smaller.

One might argue that the fix that we can apply is to require real-time video applications adjust to the new bit rate more promptly than 6.7 seconds, and disregard network elements from reducing their rate limits to less than a 1/10th within 67 seconds.

But what if the application is a chunk-based video player (e.g., HLS)? If each chunk is 10 seconds or longer, they need more than 6.7 seconds to adopt. IIUC, we allow network elements set their SCONE rate limits equal to what is indicated by congestion signals.

And separately, we should also think about the effects of SCONE packets getting lost.

These points leads me to to argue that the best solution here is to change the misuse detection logic inside network elements.

> I don't know how to implement:
> 
>> IMO, the correct approach for network elements is to retain old limits when the rate limit is reduced, but apply the new rate limits immediately when they are increased.

One possible approach is to require network elements to retain the following two objects:
1. A sliding window that can used to determine total traffic within the last 67 seconds.
2. A three- or four-entry FIFO queue that remembers the most recent rate limits being sent. As rate limits are updated no more than every 20-30 seconds, three- or four-entry queue is sufficient to remember all the rate limits being sent within 67 seconds.

Network elements never reset the sliding window (1). They simply check the conformance by seeing if total traffic within the last 67 seconds is no greater the _maximum_ rate limit retained in the FIFO queue (2) multiplied by 67.

huitema:
@martinthomson I think we need the "batch of 67 seconds worth of packet" issue right on. I agree that in practice, other mechanisms may come in play, such as congestion control, access rate limits, or the limited size of intermediate buffers. But the most shocking issue is the idea that if the network element advertises a lower rate, it must entirely thrown out the per-tuple state. It is a logical design, because saying otherwise requires us to delineate a more complex model. At the very moment when the network element wants applications to slowdown, it gives them more credits!

I don't really want to specify a pacing rate. We have consensus that we should only specify the behavior of the network elements, not that of the client. But we can have the equivalent of a pacing function if we specify that the server must *only keep a fraction* of the client state. Suppose we want a pacing rate Y and the advice arrives T seconds into the pacing period. The network element could assume that a could compute a credit equal to the lesser of `advice*67` and `Y*advice*T`. If the accumulated consumption is lower than that, the entire client state is wiped; otherwise, it is reduced by the credit.

Something like that will force the clients to pace their traffic somewhat, instead of sending series of batches. But OK, maybe that's too complicated. Maybe congestion control forced the client to only send a fraction of the "maximum batch", and so wiping the all state is acceptable. Fine, but need to explain that.

martinthomson:
> At the very moment when the network element wants applications to slowdown, it gives them more credits!

I was not happy with this either, until I reminded myself that this is a policy signal, not a congestion control signal.  If the network has the capacity, then the choice to change its policy might have a temporary effect on load (not a huge one in the grand scheme of things, to be clear, even if there are bursty senders).  But congestion controllers should protect the infrastructure.

We already talk about the interaction between this and congestion control and make it very clear that a congestion controller remains an essential part of the protocol.  https://ietf-wg-scone.github.io/scone/draft-ietf-scone-protocol.html#section-8.3

huitema (approved):
OK. I am not happy with the handling of bursting, but it is fine to have that discussion in a separate issue.

draft-ietf-scone-protocol.md
```diff
@@ -345,6 +348,27 @@ and the corresponding bitrate for each.
 {: #ex-rates title="Examples of SCONE signals and corresponding rates"}
 
 
+
+## Monitoring Period {#time}
+
+The time over which throughput advice applies is defined to be
+a period of 67 seconds.
```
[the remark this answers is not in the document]
zaheduzzaman: Yes, I think that needs some explanation.

martinthomson:
I've attempted to resolve outstanding issues here, modulo the ongoing discussion in #65, which remains an open issue for some people.  We'll continue that discussion there.

draft-ietf-scone-protocol.md
```diff
@@ -345,6 +348,38 @@ and the corresponding bitrate for each.
...
+applications are better able recover
+without exceeding limits for a significant time
+if they have exceeded limits.
+
+The choice of 67 seconds, as a prime number,
+also ensures that problems in tuning applications
+are not synchronized with other periodic effects.
+Any repeating phenomenon at this interval is therefore
+unlikely to be due to other periodic effects.
+
```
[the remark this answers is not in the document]
martinthomson: I don't like saying "we did this because that was the consensus outcome".  I prefer to use the reasons we used to decide.

Your point about seconds vs. milliseconds is well-taken.  I'll include the video segment length/key frame interval as an example.

draft-ietf-scone-protocol.md
```diff
@@ -354,7 +354,14 @@ and the corresponding bitrate for each.
...
-This value aims to strike a balance between extremes.
+Protocol participants can use a different period,
+depending on their role.
+Senders can limit their send rate over any time period
+up to 67 seconds.
+Network elements can monitor and apply limits to send rates
+using time period of at least 67 seconds.
+
+The choice of 67 seconds aims is a compromise between competing interests.
 Longer periods allow applications more flexibility
```
huitema, line 365: Are you sure about aims in that sentence?

draft-ietf-scone-protocol.md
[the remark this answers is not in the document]
huitema: The new text is improved, thanks.

draft-ietf-scone-protocol.md
```diff
@@ -365,17 +372,13 @@ if they have exceeded limits.
 
 The choice of 67 seconds, as a prime number,
 also ensures that problems in tuning applications
-are not synchronized with other periodic effects.
+are not synchronized with other periodic effects
+that might be measured in whole seconds.
+This includes segment length or key frame intervals in video applications,
+but also includes timers for middleboxes; see {{Section 4.3 of ?RFC4787}}.
 Any repeating phenomenon at this interval is therefore
 unlikely to be due to other periodic effects.
```
huitema, line 380: The sentence in the middle is much improved, but I don't fancy much the undefined "this" in the last sentence. Matter of taste, but I would write "at a 67 seconds interval".

And just to be facetious, why not 61, or 71, or 73?

martinthomson: There's no real reason for it to be 67 over any of those.  Also, 68 probably wouldn't be so bad either.

draft-ietf-scone-protocol.md
```diff
@@ -345,6 +348,41 @@ and the corresponding bitrate for each.
...
+a period of 67 seconds.
+
+Protocol participants can use a different period,
+depending on their role.
+Senders can limit their send rate over any time period
+up to 67 seconds.
+Network elements can monitor and apply limits to send rates
+using time period of at least 67 seconds.
+
+The choice of 67 seconds aims is a compromise between competing interests.
```
martinthomson, line 364:
```suggestion
The choice of 67 seconds is a compromise between competing interests.
```

draft-ietf-scone-protocol.md
```diff
@@ -345,6 +348,41 @@ and the corresponding bitrate for each.
...
+without exceeding limits for a significant time
+if they have exceeded limits.
+
+The choice of 67 seconds, as a prime number,
+also ensures that problems in tuning applications
+are not synchronized with other periodic effects
+that might be measured in whole seconds.
+This includes segment length or key frame intervals in video applications,
+but also includes timers for middleboxes; see {{Section 4.3 of ?RFC4787}}.
+Any repeating phenomenon at this interval is therefore
```
martinthomson, line 379:
```suggestion
Any repeating phenomenon at a 67 second interval is therefore
```

ihlar:
> I've heard a concern that some network elements do not want to update the packets every monitoring period if they have previously sent a rate, and it hasn't changed. This PR plus the Unknown code point in #63 doesn't quite allow for that.

It would be good to understand more about these concerns. Updating a few SCONE packets per monitoring period has close to zero cost. Is it that these elements do not want to keep per-flow state?

draft-ietf-scone-protocol.md
```diff
@@ -345,6 +348,41 @@ and the corresponding bitrate for each.
...
+Longer periods allow applications more flexibility
+in terms of how to allocate bandwidth over time.
+Shorter periods allow networks to administer policies more tightly.
+Also, when circumstances change,
+applications are better able recover
+without exceeding limits for a significant time
+if they have exceeded limits.
+
+The choice of 67 seconds, as a prime number,
+also ensures that problems in tuning applications
```
gorryfair, line 374: Does it "ensure" this (guarantee)? I wonder if it mitigates potential problems....

ihlar: That might be a bit too strong language, yes.

draft-ietf-scone-protocol.md
```diff
@@ -345,6 +348,41 @@ and the corresponding bitrate for each.
...
+are not synchronized with other periodic effects
+that might be measured in whole seconds.
```
ihlar, line 376:
```suggestion
The choice of 67 seconds, as a prime number,
also helps avoid synchronization with other periodic
effects that are commonly measured in whole seconds.
```

ihlar:
At the interim it was agreed to merge this PR as is, and move the discussion on having to update SCONE packets each window to a separate issue.

alvestrand:
> At the interim it was agreed to merge this PR as is, and move the discussion on having to update SCONE packets each window to a separate issue.

Is that issue #43 , or are you intending to open a new one?

ihlar:
> Is that issue #43 , or are you intending to open a new one?

It would be a new issue in this case.

PR #63, opened 2025-08-26
127 means unknown

### LINKED ISSUE #18: Confusion between "No limit" and "TRONE not supported on path"
See discussion on #12 

The draft currently conflates two distinct conditions in signal value 63 to mean either:
- The network can support unlimited bandwidth (higher than max range)
- TRONE is not supported by network elements on path

This creates ambiguity for applications interpreting signal 63 - should they assume unlimited bandwidth or cautiously rely on congestion control? We could disambiguate this by using different signal values:

- 62: No limit (bandwidth exceeds maximum range)
- 63: TRONE not supported on path

Or clearly specify how applications should interpret signal 63, acknowledging its dual meaning requires caution.

martinthomson:
The effect is the same as meaning "no limit" under the rules in #61, but this is clearer and seems to be what people generally want.

Closes #18.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -307,10 +307,11 @@ A Rate Signal is a 7-bit unsigned integer (0-127). The high six bits are the
 Rate Signal High Bits, and the least significant bit is the most significant
 bit of the Version field.
 
-When sent by a QUIC endpoint, the Rate Signal is set to 127, to indicate that
-no rate limit is in place or that the SCONE protocol is not supported by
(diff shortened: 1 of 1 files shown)


martinduke (approved):
Combined with the timing rules in the other PR, I believe this is robust to the multi-NE case that I was so worried about in today's meeting.

kazuho (approved):

ihlar (approved):

huitema (approved):
Yes. I think we have consensus on that.

PR #71, opened 2025-09-30
instant bitrate adaptation

### LINKED ISSUE #68: instant bitrate adaptation
One of the goals of SCONE is, if not the primary goal, is to improve user experience of video applications.

For optimal user experience, it'd be helpful to video applications if the "correct" achievable bitrate is provided the moment the connection is established.

However, in #61, we added the following statement.
> To that end, network elements MUST only update the content of datagrams on a given address tuple a few times each monitoring period. Network elements MAY update more often immediately after a change in their throughput advice, to reduce the reaction time from senders.

It means that the endpoints might not receive the correct bitrate advice until tens of seconds after the connection is established.

Are we fine with that?

If not, I think we should do something along the lines of:
* a) recommend network elements to update the first few SCONE packets on each flow, and/or
* b) allow network elements always update SCONE packets on a flow that is known to use QUIC (e.g., UDP port 443 which is a well-known port).

kazuho:
As discussed in #68, it would be helpful to the applications if the bitrate advice is available right after the handshake.

To achieve that, this pull request encourages endpoints to send a small number of SCONE packets ASAP and network elements to update them.

For the advice to endpoints, I've removed redundant text that existed elsewhere (namely section 5 and 5.3), adopting MUST from the sentence removed.

For the advice to network elements, I avoided use of RFC2119 verbs, following the previous paragraph.

Closes #68.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -721,7 +721,7 @@ signals can send SCONE packets at any time.  This is a decision that a sender
 makes when constructing datagrams.
 
 When sending SCONE packets, enpoints MUST include the SCONE packet as the first
-packet in a datagram.
+packet in a datagram, coalesced with additional packets.
 
 Upon confirmation that the peer is willing to receive SCONE packets, an endpoint
 SHOULD include SCONE packets in the first few UDP datagrams that it sends. Doing


huitema (approved):
I like the current text. Basically:

1- When sending SCONE packets, endpoints MUST send them as the first coalesced packet in the UDP payload.

2- Endpoints SHOULD send SCONE packets often enough -- unless they want to send them more often than that for some reason. Or less often. Basically their choice, because it is a trade off between overhead of SCONE and risks of not getting rate advice soon enough.

PR #81, opened 2025-10-21
Reword to emphasize "advice"

### LINKED ISSUE #78: Rate limit v. advice :-)
I think both the current protocol spec and the candidate for the applicability I-D, assert that the scope is a network that has a policy that "limits" suggesting this is intentional, rather than experienced. To me, at least as an individual, this is important. There are cases where I expect the rate is actually applied as a policy, but  also I see places where the rate might be a result of other system/propagation constraints. In both cases, the rate becomes limited, and is still experienced. I suggest we could be more careful about how we present this in both drafts. Please see examples below:

/Many access networks apply rate limits to constrain the data rate of attached devices./

/Having the network indicate what its rate limiting policy is, in a
   way that is accessible to endpoints, allows applications to use this
   information when adapting their send rate./
- I'd agree, but we could also articulate a case where there is knowledge of the upper rate and simply communicate this advice to the endpoints.

I think this could use better words:
/throughput that is different from the signaled rate limit/
- The signalled throughput advice? or rate signal?
/element that indicates a rate limit/
- The signalled throughput advice? or rate signal?

Shouldn't 3.2 also provide an example of sharing, where capacity aka rate becomes shared with other flows?  and with cross-traffic?

/prove that the throughput that is  indicated/
- The *signalled throughput advice*?

/data flowing in the same scope/
- what is the word scope here?

/network element detects a flow using more bandwidth/
- Is that capacity? (or just for consistency here it could be perhaps even mention throughput?)

5.1 Is a rate signal the same as a throughput guidance, please just explain.

/indicating no rate limit is in place/
- I am not sure that actually does indicate this. Of course, there is some sort of a rate limit somewhere along the path, but I think this text means there is no advertised throughput advice?

martinthomson:
This takes the suggestions @gorryfair made and does a little extra rewording of surrounding text to make it clearer that this is just advice.

Closes #78.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -158,13 +158,13 @@ strictly advisory.
 
 Throughput advice signals are not a substitute for congestion feedback.  Congestion
 signals, such as acknowledgments, provide information on loss, delay, or ECN
-markings {{?ECN=RFC3168}} that indicate the real-time condition of a network
-path.  Congestion signals might indicate a throughput that is different from the
-signaled rate limit.
+markings {{?ECN=RFC3168}} that indicate the real-time condition of a network path.
+Congestion signals might indicate a throughput limit
+that is different from the signaled throughput advice.
 
 Endpoints cannot assume that a signaled rate limit is achievable if congestion
 signals indicate otherwise.  Congestion could be experienced at a different
-point on the network path than the network element that indicates a rate limit.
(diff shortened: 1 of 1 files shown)


ihlar (approved):

martinthomson:
@huitema, I contemplated opening an issue for the apparent conflict.  I didn't in the end, mostly because I see this as an essential tension in the design.

I just wanted to flag this in case you wanted to open that issue.

Issue #84, opened 2025-10-23
"maximum achievable throughput" or "maximum sustained throughput" ?


zaheduzzaman:
I noticed that the draft now uses both “maximum achievable throughput” and “maximum sustained throughput” interchangeably, however, they are two different concepts. 

From my understanding, what the SCONE network elements can provide is the maximum achievable throughput, They can’t guarantee what is sustained for a particular 4-tuple (for instance when a number of flows share the same max).

Unless we define what "sustrained" means, using both terms are confusing. I suggest we use “maximum achievable throughput”.

martinthomson:
Want to write up a PR to squash those few instances of "sustained" you found?

kazuho:
> To my understanding, the "max sustrained throughput" is something that can be maintained over certain period of time. Here, there is some sort of gurantee to provide that throughput over a period of time, for scone we don't know that configured period. However, we have 67s of monitoring period.

I think the most definition at the moment is the one we have in [Section 5.4](https://ietf-wg-scone.github.io/scone/draft-ietf-scone-protocol.html#section-5.4-2): i.e., _an endpoint that receives throughput advice might receive multiple different rate limits. If advice is applied by applications, applications MUST apply the lowest throughput advice received during any monitoring period_?

Reading the quoted paragraph, I realize that we also use the term "throughput advice" as well.

Regarding the signal that SCONE provides, I'm starting to wonder if "throughput advice" is a term better than "rate limit." The problem with "rate limit" is that, in the document, we talk about things like "rate limits" or "rate limit policy" to refer to CC-based throttling techniques not necessary tied to SCONE.

PR #88, opened 2025-11-02
Switch terminology

martinthomson:
As we've discussed on #84, we used "rate limit" too often.  Fix that.


kazuho (commented):
Thank you for doing this!

Do we want to rename "Rate Signal" to "Throughput Signal" (or advice) too?

huitema (approved):
Yes, standardizing on "throughput advice" does clarify and improve the draft.

PR #93, opened 2025-11-26
What advice means

martinthomson:
For a sender, advice is cumulative.  They should follow any advice they have received in the last monitoring period.

For a network, the advice they send is something that might take time to respond to.  They should allow a monitoring period before they expect an application to be able to follow advice.

The text on following/monitoring advice is now shorter.

I've also updated the sample code to track all active advice.  If multiple pieces of advice are received, this now tracks sending against all of them.  A sender/application follows the least permissive; a network allows the most permissive.

The effect of this is similar to what was in the previous sample code, but it's a little less vulnerable to weirdness.

For instance, if the advice increases, the old send rate is monitored until that advice expires.  Applications will therefore continue to follow that old advice, but switch to the new one once the old advice is more than 67s old.

Similarly, if the advice to an application reduces, the old rate is also monitored, but the new rate also applies.  Because the new rate is initialized with no memory of what has been sent, the new rate probably allows for more sending, but because the old advice remains active until it expires, you don't get a complete reset.

The cost is that the code tracks as many rates as there is different advice received over a monitoring period.

I'm not sure that we really need the sample code here, but it was a useful exercise to go through in order to understand how this might work in practice.  Looking for input on whether it is useful to retain.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -657,6 +657,10 @@ and adapt their sending rate.
 
 Any monitoring and policy enforcement could be implemented
 in different network elements than the ones that signal throughput advice.
+However, network elements MUST NOT enforce throughput based on throughput advice
+found in SCONE packets received from other entities, because unlike endpoints,
+network elements do not have the capability to validate other QUIC packets
+contained in the same datagram; see {{fake-packets}}.
 
 
 ## Flows That Exceed Throughput Advice {#policing}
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-protocol.md
```diff
@@ -451,8 +451,9 @@ and might include other, application-specific constraints.
 Applications MAY discard throughput usage state
 when they receive throughput advice that indicates a reduced rate.
```
kazuho: sounan?

言いたいことは、受け取ったadviceにmonitoring_period内にadaptしたいといけない、じゃないのかな

Issue #75, opened 2025-10-15
Is sending advice every 67 seconds too computationally expensive for network elements?


afrind:
This discussion arose from #70 during the 10/15 interim.

Some network elements are already deployed with relatively fixed capacity.  SCONE's current requirement is that in order to maintain a limited flow, the NE needs to mark packets every monitoring period.  @ihlar made some back of the envelope [computations](https://github.com/ietf-wg-scone/scone/pull/70#discussion_r2433930965) that make it seem like it's ok.  This is a placeholder issue we can revisit once we can measure the impact in production(like) requirements.

martinthomson:
As @ihlar noted in that issue, this is pretty cheap.

I did a rough implementation of the core function in rust, which you can see [here on the compiler explorer](https://rust.godbolt.org/z/rh385f49v).  In that approach, every packet you pass will cost 14 instructions (which should all be low latency).  Updating costs 6 more.  To put it in perspective, in software, the additional check adds 28ns on a very low-end CPU, the update will like cause a pipeline stall, but those are still only a few tens of nanoseconds.

irfan-ali-ia:
With SCONE packet it is efficient to identify the field where the the rate signal is stored. However, modifying the value does have bigger compute implications (eg. also computed checksum over the entire UPD datagram), specially for gateways to cellular networks (UPF) which are processing millions of flows and performing other function on these flows. It all comes down to capacity/CPU. So for systems where the rate guidance does not change often, if the server provides an indication that it had received rate guidance (either rate or some function of the rate relayed by the client), and that it is keeping soft-state of  the guidance (which could be for much longer than 67 seconds, eg. server knows that the particular cellular network does not change guidance often), the UPF can modify rate either (a) when it gets a signal that the server's soft-state has expired, or (b) when rate guidance needs to be updated, eg due to policy change. This would also mean that we do not have to make assumptions of 67s soft-state, which though a nice prime number, may not be applicable to the system in question.

alvestrand:
Incrementally updating the UDP checksum is described in https://datatracker.ietf.org/doc/html/rfc1624.

In trying to estimate the cost of updating the rate guidance for one million flows every 67 seconds, if I assume 10 instructions, in cache, on a CPU with 1 GHz instruction rate, the update of one million flows can be computed in 1/100 of a second, if my back-of-the-envelope calculation is accurate, leaving 66.99 seconds of the CPU available for other purposes.

[...]

PR #95, opened 2025-11-26
Monitoring is a network policy choice

### LINKED ISSUE #91: Confusion stemming from the responsibility of the element
For SCONE to work as a protocol a Network Element (or the "SCONE Function" as we might call it) minimally has to periodically modify SCONE packets. In the Deployment section we currently say:

> Network elements can monitor flows to determine which flows are from applications that respect throughput advice. This section outlines an exemplary algorithm for network elements.

> This monitoring algorithm is guidance only. Network elements are advised that monitoring any more strictly than the following likely invalidates their advice. Network elements could choose not to monitor in this way or use a looser monitoring approach. For instance, monitoring over a longer time window than the monitoring period (67s) or using a higher rate than is signaled is possible.

This does imply some degree of it being an "optional" responsibility, but we have gotten feedback from those going in "cold" to the document that the monitoring feels much more mandatory. Most of the subsequent discussion resolve around monitoring and not the actual protocol.

I think we should consider structural changes to the document (e.g. moving example monitoring to an addendum, making the language more explicit about how it's not normative, moving to another document entirely) to address this.

martinthomson:
This builds on top of #93.  It probably should be merged in with it.

It attempts to address #91 by significantly softening the language around monitoring.

@mjoras does this help?

Closes #91.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -659,8 +659,9 @@ was followed.
 A network element SHOULD base its monitoring
 on the maximum value that was configured to apply
 during the preceding two monitoring periods.
-A longer period might be used
-to account for the possibility that SCONE packets are lost.
+If the network element cannot update the throughput advice in every SCONE packet
+(or can do so only infrequently), a longer period might be used
+to account for the possibility that the updated SCONE packets are lost.
 This allows applications time to receive advice
 and adapt their sending rate.
 


martinthomson:
I've also added a note that points out that updating SCONE packets, monitoring, and policy enforcement are separate functions that don't need to be co-located.  (I hope that was obvious, but it's now also explicit.)

draft-ietf-scone-protocol.md
```diff
@@ -612,78 +612,81 @@ To reduce the risk of synchronization across multiple senders,
...
-
-* the current throughput advice,
-* any state necessary to monitor throughput
-  (that is, throughput usage state, such as the state in {{sliding-window}}),
-* a timer used for rate increases, and
-* the time at which that throughput advice was last updated.
+or using a higher rate than is signaled,
+has no risk of incorrect classification.
 
+Network elements need to wait until advice is signaled.
```
afrind: This reads more like a section title, since the next sentence is describing the waiting too?

draft-ietf-scone-protocol.md
```diff
@@ -612,78 +612,81 @@ To reduce the risk of synchronization across multiple senders,
...
 When advice is set or updated,
-the network element waits until
-it receives the next SCONE packet on affected flows.
-It then updates the throughput advice in that packet ({{apply}}).
+monitoring cannot start until the network element
+receives a SCONE packet
+and updates the throughput advice in that packet; see {{apply}}.
```
afrind: If an element observes advice in a SCONE that is limited - either by the original sender or an upstream element - and does not update it, is it permitted to monitor?

draft-ietf-scone-protocol.md
```diff
@@ -612,78 +612,80 @@ To reduce the risk of synchronization across multiple senders,
...
+or using a higher rate than is signaled,
+has no risk of incorrect classification.
 
 When advice is set or updated,
-the network element waits until
-it receives the next SCONE packet on affected flows.
-It then updates the throughput advice in that packet ({{apply}}).
+monitoring cannot start until the network element
+receives a SCONE packet
+and updates the throughput advice in that packet; see {{apply}}.
```
```suggestion
transmits a SCONE packet with
throughput advice other than `Unknown` (127); see {{apply}}.
```

draft-ietf-scone-protocol.md
```diff
@@ -612,78 +612,81 @@ To reduce the risk of synchronization across multiple senders,
...
-* the time at which that throughput advice was last updated.
+or using a higher rate than is signaled,
+has no risk of incorrect classification.
 
+Network elements need to wait until advice is signaled.
 When advice is set or updated,
-the network element waits until
-it receives the next SCONE packet on affected flows.
-It then updates the throughput advice in that packet ({{apply}}).
+monitoring cannot start until the network element
```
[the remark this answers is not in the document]
martinthomson: Thanks.

draft-ietf-scone-protocol.md
```diff
@@ -612,78 +612,80 @@ To reduce the risk of synchronization across multiple senders,
...
+or using a higher rate than is signaled,
+has no risk of incorrect classification.
 
 When advice is set or updated,
-the network element waits until
-it receives the next SCONE packet on affected flows.
-It then updates the throughput advice in that packet ({{apply}}).
+monitoring cannot start until the network element
+transmits a SCONE packet with
+throughput advice other than `Unknown` (127); see {{apply}}.
```
kazuho: This paragraph sounds like monitoring devices might throttle flows that don't follow throughput advice from any network device, _including network devices on other hops_. Is that intentional?

I get the intent to describe the algorithm based on what the signaled advice is (because that is what the applications would use in order to comply), however my concern is that if we are to consider such deployments, we might need to expand security considerations how an on-path attacker might abuse SCONE to attack monitoring devices deployed by in other hops.

kazuho: > We can't prevent a network element from monitoring the advice that is contained in packets it's sending,

This is true.

> we can't distinguish between "passing through" advice from upstream and "it set the same rate I was going to set", can we?

From the on-wire signal alone, we can't distinguish those cases, agreed.

However, that is exactly why I think the PR is problematic in its current state.

The phrase "transmits a SCONE packet with 〜" can naturally be read as covering a SCONE signal that originated upstream and is merely being passed through unchanged. And in that case, the following proposed normative text:

> "A network element SHOULD base its monitoring on the maximum value it signaled for throughput advice over at least the preceding monitoring period."

can also be read as telling a monitoring/enforcement device to treat such pass-through / third-party advice as the basis of its monitoring, and by extension, policing / enforcement decisions.

That model does not match the current security model described in the draft, where devices that _act_ on SCONE advice (endpoints) only do so when the SCONE signal is attached to a valid QUIC packet, preventing an off-path attacker from inserting spurious advice. A network element that relies on pass-through / third-party advice cannot apply that "accepts only SCONE signals attached to valid QUIC packets" condition, so text that assumes or endorses that model should not be included in the I-D.

To conclude, I think the draft should talk about the advice that the _hop intends to apply locally,_ rather than whatever SCONE signal that happens to be present in transmitted packets.

huitema: Relying on packets that "just happen to pass by" is a security breach waiting to happen. The network element has no way to verify that the Scone packet that it sees is genuine. For all we know, an adversary may create a fake Scone packet, coalesced with a fake 1RTT packet. The network element sees that and start its monitoring clock, but the endpoint recognizes the packet as fake and does not heed it. Conflict ensues.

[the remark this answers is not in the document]
martinthomson: We're getting into the meat of this now.

This is the problem: if the network has a policy, then it might suspend enforcement of that policy if it gets a chance to use SCONE.  But we don't need to say that in this particular way.  Let me try again...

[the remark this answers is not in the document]
kazuho: I like this :+1:

Maybe we can tighten the first sentence to: When a network changes the advice, applications need time to adjust their sending behavior.

By omitting "it gives," we avoid the impression that the change starts from when the SCONE packet is sent.

draft-ietf-scone-protocol.md
```diff
@@ -612,78 +612,80 @@ To reduce the risk of synchronization across multiple senders,
...
+
+Networks can monitor flows to determine if applications follow advice;
+see {{monitoring}}.
+A network could choose to either disable or loosen policy enforcement
+for flows where SCONE is active,
+but re-enable or tighten enforcement if monitoring indicates
+that throughput advice is not being respected.
+
+Any monitoring and policy enforcement could be implemented
+in different network elements than the ones that signal throughput advice.
```
kazuho: It might make sense to change like below, to avoid the impression that monitoring and enforcement can be done base on untrusted rate signals (#98):
```suggestion
Monitoring and policy enforcement based on network policy could be implemented
in different network elements than the ones that signal throughput advice.
```

Though, separately, I think this paragraph might need to be moved to #monitoring.

draft-ietf-scone-protocol.md
```diff
@@ -612,78 +612,80 @@ To reduce the risk of synchronization across multiple senders,
...
+throughput advice other than `Unknown` (127); see {{apply}}.
 
 A network element can then monitor affected flows
 to determine whether the throughput advice it provided
 was followed.
 
 A network element SHOULD base its monitoring
-on the maximum value it provided for throughput advice
+on the maximum value it signaled for throughput advice
 over at least the preceding monitoring period.
```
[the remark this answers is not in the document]
martinthomson:
```suggestion
on the maximum value that was configured to apply
during the preceding two monitoring periods.
```
?

huitema (approved):
I like the changes since my previous review, I think we have now something quite clear.

draft-ietf-scone-protocol.md
```diff
@@ -612,78 +612,83 @@ To reduce the risk of synchronization across multiple senders,
...
 
 A network element SHOULD base its monitoring
-on the maximum value it provided for throughput advice
-over at least the preceding monitoring period.
-Additional time might be allowed
-to account for the possibility that some SCONE packets might be lost.
+on the maximum value that was configured to apply
+during the preceding two monitoring periods.
+A longer period might be used
+to account for the possibility that SCONE packets are lost.
```
kazuho: IIRC, we assume 3 SCONE packets to be sent within one monitoring period. Therefore, it seems to me that this advice is not needed for network elements capable of updating every SCONE packet that they forward.

Assuming that is the case, I think we can drop this sentence, or change to the following if we are to close #75 along the way.

```suggestion
If the network element cannot update the throughput advice in every SCONE packet
(or can do so only infrequently), a longer period might be used
to account for the possibility that the updated SCONE packets are lost.
```

PR #96, opened 2025-11-26
Update intro

### LINKED ISSUE #69: Definition of "Network Elements" is missing.
Current text uses the term "network element" but it does not define or describe the network elements. It appears that there are certain expectations from a network element -

- It sits in the access network.
- It has visibility to UDP 4-Tuple flow sand it maintains some sort of states about the rate the individual flows are using.
- It has rate limiting policies or have access to it.
- It can detect SCONE packets in those UDP 4-tuple flows and put rate advices in those packets.

This all I am infering from the text of this draft, but, I am not sure if everyone would endup infering the same.

What I infer makes the "network element" a bit unique, that not any router or any device can operate the SCONE protocol. Hence, the document should clearly define this "network element".

This would also help other discussions regarding manageability and applicablity discussions over different access networks as it would help to understand what are those access network there we can have such network elements.

martinthomson:
This attempts to shorten the intro by moving some of the text that has been added over time to the applicability section.

It also clarifies that "network element" is any function that is on-path.  I'll add the other thing we agreed to at the last meeting to another PR that touches the relevant text.

Closes #83.
Closes #69.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -165,8 +165,8 @@ point on the network path than the network element that signals throughput advic
 Therefore, endpoints need to respect the send rate constraints that are set by a
 congestion controller.
 
-Networks are not limited to using SCONE to communicate information
-about rate limiting policies.
+Networks can use SCONE to communicate throughput advice
+for reasons other than rate limiting policies.
 For example, a network element in an access network
 could provide throughput advice to guide application use of network capacity,
 in any way that is separate from any signals


huitema (approved):
Looks good.

draft-ietf-scone-protocol.md
```diff
@@ -168,6 +165,20 @@ point on the network path than the network element that signals throughput advic
 Therefore, endpoints need to respect the send rate constraints that are set by a
 congestion controller.
 
+Networks are not limited to using SCONE to communicate information
+about rate limiting policies.
+For example, a network element in an access network
+could provide throughput advice to guide application use of network capacity,
+in any way that is separate from any signals
+that are intended to influence congestion response.
```
afrind: I see this is just moved test, but...

This document is describing SCONE but this reads like placing limits on what you can do if you are not SCONE. Maybe just "could provide throughput advice...through out-of-band mechanisms".

draft-ietf-scone-protocol.md
```diff
@@ -168,6 +165,20 @@ point on the network path than the network element that signals throughput advic
...
+
+Throughput advice indicate temporary increases in available capacity,
```
```suggestion
Throughput advice can indicate temporary increases in available capacity,
```
?

draft-ietf-scone-protocol.md
```diff
@@ -168,6 +165,20 @@ point on the network path than the network element that signals throughput advic
 Therefore, endpoints need to respect the send rate constraints that are set by a
 congestion controller.
 
+Networks are not limited to using SCONE to communicate information
+about rate limiting policies.
```
```suggestion
Networks can use SCONE to communicate throughput advice
for reasons other than rate limiting policies.
```

draft-ietf-scone-protocol.md
afrind: Yes, thanks

PR #99, opened 2025-12-03
forbid enforcement by network elements based on 3rd party signals

### LINKED ISSUE #98: network elements must not use the bitrate field set by third party
Network elements must not use (e.g., for monitoring, throttling) the rate signal included in the SCONE packets, unless they have a trust relationship with the party that set the rate signal included in _that_ SCONE packet.

Quoting from https://github.com/ietf-wg-scone/scone/pull/95#discussion_r2567231140:
> (doing so) not match the current security model described in the draft, where devices (endpoints) that act on SCONE advice only do so when the SCONE signal is attached to a valid QUIC packet, preventing an off-path attacker from inserting spurious advice. A network element that relies on pass-through / third-party advice cannot apply that "accepts only SCONE signals attached to valid QUIC packets" condition, so text that assumes or endorses that model should not be included in the I-D.

and from https://github.com/ietf-wg-scone/scone/pull/95#discussion_r2567256749:
> Relying on packets that "just happen to pass by" is a security breach waiting to happen.

Aside from describing how monitoring could be done (which IIUC is the aim of #95), it might make sense to have a dedicated issue for discussing if we need a statement forbidding / discouraging such behavior.

kazuho:
This PR closes #98; it builds on top of #93, assuming that #93 will be merged.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -657,6 +657,10 @@ and adapt their sending rate.
 
 Any monitoring and policy enforcement could be implemented
 in different network elements than the ones that signal throughput advice.
+However, network elements MUST NOT enforce throughput based on throughput advice
+found in SCONE packets received from other entities, because unlike endpoints,
+network elements do not have the capability to validate other QUIC packets
+contained in the same datagram; see {{fake-packets}}.
 
 
 ## Flows That Exceed Throughput Advice {#policing}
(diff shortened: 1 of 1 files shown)


huitema (commented):
Of course, every defense as a counter attack. Suppose a malicious endpoint that does not want to be see its traffic slowed down. It will initiate SCONE packets in which, in violation to the spec, it sets the advice data to a low value. The network elements see these packets, and will delay enforcement because they cannot set the advice.

But that's probably silly. The malicious endpoint could also not send any SCONE packet at all.

Which means we have to give some lee-way to the network elements. If they could not find a packet that the can overwrite for a whole period, and they see packets with a lower limit, they should probably just start enforcing.

afrind:
I noticed this got merged into a branch that had already merged into main, I don't think it's in main yet.

Issue #106, opened 2025-12-08
Please explain a receiver can control the release of flow control credit.


gorryfair:
OLD:
Alternatively, a receiver can control the release of flow control credit (see [Section 4](https://rfc-editor.org/rfc/rfc9000#section-4) of [[QUIC](https://ietf-wg-scone.github.io/scone/draft-ietf-scone-protocol.html#QUIC)]) to indirectly limit the sending rate of a peer.

- I saw this, and couldn't argue that a design could so this. But, there again, there are many things a receiver could do. If this is being called out as one of these why is this a good or bad idea to do this? If this is a recommended approach, I'd expect an example and explanation to avoid reacting to the signal at multiple levels in the stack, which I suspect could be awkard when there is a changing signal. Who has experience? Are there thoughts?

martinthomson:
I'm not a believer in the use of flow control for this, but I know that some people are actively doing that.

At a high-level, it isn't especially complicated.  If you want to receive at a rate of X bytes/second, you limit the rate at which you provide flow control to X bytes/second.

Of course, this inevitably leads to something less than X bytes/second of throughput because any signaling needs to propagate.  So documenting how you avoid that happening is the subject of an entirely new internet draft.  (I know that some people are working on documenting flow control auto-tuning algorithms, which turns out to be subtle.)

I don't know how to address this.  A simple way to manage this would be to cut the text, but that seems unsatisfactory.  I'm happy with where it stands, even though this is suggesting something that is probably too subtle for many implementers to get right.  But, given that good flow control algorithms are not subject to standardization, I'm comfortable with that remaining as "special sauce".

gorryfair:
The above make sense at least to me, and splitting into 2 para's help!

PR #112, opened 2025-12-09
Gorry's review

martinthomson:
This includes a bunch of commits, each addressing a different aspect of Gorry's review.  Merging this should close all of the issues that @gorryfair raised, aside from #106.  My inclination there is to do nothing, but it's more complicated than that.  See that issue discussion.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -223,8 +223,8 @@ to be split between multiple active flows.
 The same UDP address tuple might be used for multiple QUIC connections.  A
 single signal might be lost or only reach a single application endpoint.
 Network elements can apply SCONE advice
-to all flows that include SCONE packets
-to ensure that advice is received by the endpoints of all flows.
+to all QUIC connections that include SCONE packets
+to ensure that advice is received by all application endpoints.
 
 The signaled advice applies to the flow of packets
 on the same UDP address tuple for the duration of


draft-ietf-scone-protocol.md
```diff
@@ -261,10 +286,12 @@ limiting is necessary.  Alternatively, a receiver can control the release of
...
-Some applications offer options for rate control that can offer superior
-outcomes.  Real-time and streaming video applications are able to adjust video
-quality to fit within a target throughput.  For instance, an HTTP Live Streaming
-client {{?HLS=RFC8216}} can ask for lower bitrate, lower quality media segments.
+Some applications offer options for rate control that can offer superior outcomes.
+Applications that are able to adapt bit rates,
+such as real-time and streaming video applications,
+can adjust video quality to fit within a target throughput.
+For instance, an HTTP Live Streaming {{?HLS=RFC8216}} or DASH {{DASH}}
+client can ask for lower bitrate, lower quality media segments.
```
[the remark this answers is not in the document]
kazuho: Thanks, but as to your points:
* Regarding retransmissions, I'm not sure if it is relevant, as the problem exists with video bit rate selection. To paraphrase, SCONE assumes a low PLR.
* I wouldn't mind keeping status quo as-is, but if we want to explain why, maybe we should be more direct about the types of traffic might respect SCONE using flow control. We could use background software updates as an example, pointing out that, in some cases, avoiding CC-throttled is more important than the speed of background operation.

gorryfair (approved):

draft-ietf-scone-protocol.md
```diff
@@ -281,10 +286,12 @@ limiting is necessary.  Alternatively, a receiver can control the release of
...
 rate of a peer.
 
-Some applications offer options for rate control that can offer superior
-outcomes.  Real-time and streaming video applications are able to adjust video
-quality to fit within a target throughput.  For instance, an HTTP Live Streaming
-client {{?HLS=RFC8216}} can ask for lower bitrate, lower quality media segments.
+Some applications offer options for rate control that can offer superior outcomes.
+Applications that are able to adapt bit rates,
+such as real-time and streaming video applications,
+can adjust video quality to fit within a target throughput.
```
gorryfair: Seems slightly clunky. It could be something alonmg the lines of: "Video-based applications that can adjust video quality, such as real-time and streaming video applications,
can adjust their rate to fit within a target throughput"

gorryfair (changes requested):

gorryfair (approved):

draft-ietf-scone-protocol.md
```diff
@@ -206,24 +222,43 @@ to be split between multiple active flows.
 
 The same UDP address tuple might be used for multiple QUIC connections.  A
 single signal might be lost or only reach a single application endpoint.
-Network elements that signal about a flow might choose to send additional
-signals, using connection IDs to indicate when new connections could be
-involved.
+Network elements can apply SCONE advice
+to all flows that include SCONE packets
```
```suggestion
to all datagrams that include SCONE packets
```

Is this the intent?

draft-ietf-scone-protocol.md
```diff
@@ -601,11 +635,15 @@ if is_long and (packet_version & 0x7fffffff) == SCONE_VERSION_BITS:
...
 
-A network element needs to ensure that it sends updated rate signals
+To avoid throughput advice expiring,
+a network element needs to ensure that it sends updated rate signals
 with no more than a monitoring period ({{time}}) between each update.
 Because this depends on the availability of SCONE packets
 and packet loss can cause signals to be missed,
 network elements might need to update more often.
+Ideally, network elements update advice in SCONE packets
+twice per monitoring period,
```
```suggestion
at least twice per monitoring period,
```
More than twice is even better.

kazuho (commented):
Thank you for all the edits. The PR is looking good.

I'd be fine to discuss how to close #106 separately.

draft-ietf-scone-protocol.md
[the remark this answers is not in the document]
kazuho: Opened #113 so that we could discuss the point separately.

draft-ietf-scone-protocol.md
```diff
@@ -172,11 +183,12 @@ could provide throughput advice to guide application use of network capacity,
...
-or temporarily reduced capacity
-due to persistent overuse, equipment faults, or other transient issues.
-This applies to increases or reductions
-that are expected to last at least one minute.
+In addition to rate limiting policies,
+throughput advice can indicate temporary increases in available capacity
+or temporarily reduced capacity.
+This includes persistent overuse, equipment faults, or other transient issues.
+Providing advice is applicable if increases or reductions
+are expected to last at least one minute.
```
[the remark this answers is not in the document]
martinthomson: I think that the problem is "one", not that the text is vague.  Because what we really want to say is "at least several monitoring periods".  There's no value in signaling for anything of shorter duration.

```suggestion
are expected to last for more than one monitoring period; see {{time}}.
```

draft-ietf-scone-protocol.md
martinthomson: "flows" is correct in this context.  The title of the section is "Per-Flow Signal".

kazuho (approved):
Thank you. Looks good.

PR #113, opened 2025-12-10
explain how flow control can be used

### LINKED ISSUE #106: Please explain a receiver can control the release of flow control credit.

kazuho:
At the moment, the PR points to #112. We can either merge this PR to #112, or merge #112 and then discuss separately, by changing the base branch to main.

Closes #106.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -297,7 +297,8 @@ that the peer can adapt its sending behavior.
 
 Alternatively, when such direct rate adaptation by the sending peer is not
 available, a receiver might control the release of the flow control credit (see
-{{Section 4 of QUIC}}) to indirectly constrain the peer's sending rate. By doing
+{{Section 4 of QUIC}}) to indirectly constrain the peer's sending rate over QUIC
+streams. By doing
 so, the receiver can keep the peer's sustained sending rate within the advised
 limit and reduce the likelihood of triggering network enforcement using
 congestion signals. This approach is imprecise, as the delays in communicating


draft-ietf-scone-protocol.md
```diff
@@ -297,6 +291,19 @@ For instance, typical HTTP Live Streaming {{?HLS=RFC8216}} or DASH {{DASH}}
...
+For such applications, the most straightforward way to keep within the
+throughput advice is to communicate the advised limit to the sending peer, so
+that the peer can adapt its sending behavior.
+
+Alternatively, when such direct rate adaptation by the sending peer is not
+available, a receiver might control the release of the flow control credit (see
+{{Section 4 of QUIC}}) to indirectly constrain the peer's sending rate. By doing
+so, the receiver can keep the peer's sustained sending rate within the advised
+limit and reduce the likelihood of triggering network enforcement using
+congestion signals. This approach may reduce average throughput, but for
```
```suggestion
congestion signals. This approach can be imprecise, as the delays in communicating
flow control credit can reduce throughput well below target rates and loss recovery
can cause rates to exceed targets. For some applications, like
```

huitema (commented):
Given that the datagram hole is big enough to pass truckloads of RTP packets, maybe we should just completely abandon this discussion of using flow control to manage how quickly the peer can send...

PR #114, opened 2025-12-11
Remove mention of flow control for rate control

martinthomson:
At best, this is hard to use and indirect.  At worst, this is totally ineffectual.

@kazuho, does this work for you?

Closes #113.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -283,18 +283,26 @@ There are no guarantees that updated advice will be sent at such events.
 The SCONE throughput advice is advisory (see {{advisory-signal}}).
 Applications that chose to follow it will do so in the way that best suits their needs.
 
-The most obvious way to keep within the limits set by throughput advice is to
-inform the sending peer of the limit so that the peer can do whatever rate
-limiting is necessary.
+The most obvious way to follow throughput advice is to
+inform the sending peer of the advice so that the peer
+can adjust send rates as necessary.
+This document does not provide specific guidance on how applications
+might adapt their use of network capacity in response to advice.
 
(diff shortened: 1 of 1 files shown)


huitema (approved):
Yes. That's the simplest way to fix this, we don't have to speculate on what apps will do.

kazuho (commented):
> we don't have to speculate on what apps will do.

Yeah. The problem at the moment is that this subsection explains how things can be implemented, then goes on to clarify why some applications can do so (e.g., video players switching between segments with different bitrates). In that structure, mentioning flow control without framing it in terms of an application is not only odd, but, as @huitema notes, could also mislead readers into thinking that the technique applies equally well to non-stream transfers.

I think the subsection might flow better if we focus on applications that might benefit from SCONE, and optionally note how specifically they might use SCONE.

Given the illustrative nature of the subsection, I think it's worth pointing out that there are use cases beyond video.

Software updates are probably the largest type of traffic after video and are also spiky. User experience of other applications suffer when CC-based throttling is applied equally to both software updates and other traffic.

Maybe we can just stop there, without explaining in detail how bulk transfers might respect SCONE.

Honestly, I do not see any pitfalls in using a flow-control-based approach for bulk downloads. If the throughput advice is 1Mbps, for example, I would issue flow-control credit every 8 seconds. for 1MB beyond the last contiguous byte received). But that is just one way of doing it, and I agree that we do not want to discuss in the document how these things can be done (or should not be done).

martinthomson:
Thanks @kazuho, @ihlar.  I've made some more significant changes to the section based on the feedback here.  Let me know if that works for you.

kazuho (approved):
Thank you @martinthomson! LG.

PR #117, opened 2025-12-23
WGLC Nits

martinduke:
A few nits I found in WGLC review.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -220,14 +220,14 @@ to be split between multiple active flows.
 
 ## Per-Flow Signal
 
-The same UDP address tuple might be used for multiple QUIC connections.  A
+The same address, UDP port tuple might be used for multiple QUIC connections.  A
 single signal might be lost or only reach a single application endpoint.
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
"address tuple" was used as QUIC used it, but we can take this change as we don't need to rely on QUIC for that definition.

kazuho (approved):

draft-ietf-scone-protocol.md
```diff
@@ -220,14 +220,14 @@ to be split between multiple active flows.
...
-The same UDP address tuple might be used for multiple QUIC connections.  A
+The same address, UDP port tuple might be used for multiple QUIC connections.  A
 single signal might be lost or only reach a single application endpoint.
 Network elements can apply SCONE advice
 to all QUIC connections that include SCONE packets
 to ensure that advice is received by all application endpoints.
 
 The signaled advice applies to the flow of packets
-on the same UDP address tuple for the duration of
+on the same address, UDP port tuple for the duration of
```
```Suggestion
on the same tuple of IP version, IP address, and UDP port number for the duration of
```

PR #118, opened 2026-01-04
Tunnels

martinthomson:
This might be stating the obvious, but it's probably worthwhile.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -236,6 +236,14 @@ Rate limiting policies often apply on the level of a device or subscription,
 but endpoints cannot assume that this is the case.
 A separate signal can be sent for each flow.
 
+Network elements that apply throughput advice to a flow
+that provides an encrypted tunnel for an encapsulated flow
+(such as {{?CONNECT-UDP=RFC9298}})
+only applies to the outermost flow.
+Advice can be applied on flows that are subsequently encapsulated,
+but following that advice can have security implications;
+see {{active-attacks}}.
+
 
 ## Undirectional Signal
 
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-protocol.md
```diff
@@ -236,6 +236,14 @@ Rate limiting policies often apply on the level of a device or subscription,
...
 A separate signal can be sent for each flow.
 
+Network elements that apply throughput advice to a flow
+that provides an encrypted tunnel for an encapsulated flow
+(such as {{?CONNECT-UDP=RFC9298}})
+only applies to the outermost flow.
+Advice can be applied on flows that are subsequently encapsulated,
+but following that advice can have security implications;
+see {{active-attacks}}.
+
```
gorryfair: I'm pleased to see some thought about tunnels. The additional text seems to ONLY relate to a QUIC flow provides an encrypted tunnel, and not other encrypted tunnels. Ought this to include "QUIC" somewhere in the sentence (e.g., before "flow")?

alvestrand: I'd like to think that the tunneling advice applies to any sort of tunnel (my particular concern is with TURN flows, once (if) we get SCONE working with ICE), so I'm happy to see generic language here.

gorryfair: I am happy also to see generic advice, but for encrypted tunnels the network element can only apply the SCONE protocol to the outer-most QUIC packet header.

martinthomson: In all cases, SCONE can only (currently) be applied to QUIC flows.  Like Harald, I thought about this a fair bit and decided not to add "QUIC", both because that is addressed elsewhere and because the generic language is clear, short, and does open that possibility of having other SCONE applications.

(I like SCONE-ICE, though maybe iced scones, rather than icy scones.)

huitema (approved):
I think this text is correct. The network elements cannot see inside the tunnel, and the advice that might be given to encapsulated flows is something for the tunnel provider to decide. We could imagine all kinds of sophisticated handling. The tunnel provider is de facto a "network element", and could somehow provide SCONE advice to encapsulated flow. It could also implement some kind of AQM to indirectly provide the advice to these flows. (The RFC9298 cannot use ECN, see section 6.2 of RFC9298, so there is no need to consider that.)

PR #119, opened 2026-01-05
The text here was outdated

martinthomson:
Missed in recent updates.  Changing the title seemed like a good idea as well.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -740,7 +740,7 @@ but re-enable or tighten enforcement if monitoring indicates
 that throughput advice is not being respected.
 
 
-# Version Interaction {#version-interaction}
+# Endpoint Usage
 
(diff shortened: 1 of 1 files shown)


kazuho (approved):

huitema (approved):
I am happy to see the removal of "Graceful handling of network elements that might only recognize one version." Do we need to say something about that "partial support" in the security section? A network element or an endpoint that only supports one of the versions will probably expose itself to creative use of the other version by attackers.

martinthomson:
The less said about supporting one or other SCONE version, the better.  It's a package deal.  Support both.

The only reason I can think of for paying attention to one of the two versions is if you are just monitoring for a specific value for the advice.  Why you would do that is beyond me, though maybe the "unspecified" value might qualify as something that is worth watching for.  In any real usage, you need to monitor the version because it contains the low bit of the advice.

PR #122, opened 2026-01-07
Alan's review

martinthomson:
These are all editorial.  There were some substantive things raised in the review, which are not included here.  I think that the right response is to say "no change", but I don't want to foreclose on that.  If discussion leads to something substantive, we'll need separate changes.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -137,10 +137,10 @@ is shown in {{f-scone}}.
 QUIC endpoints that receive modified SCONE packets observe the indicated
 version, process the QUIC packet, and then record the indicated rate.
 
-Throughput advice only apply to the direction and path for which they are
+Throughput advice only applies to the direction and path for which it is
 received.  A connection that migrates or uses multipath
 {{?QUIC-MP=I-D.ietf-quic-multipath}}
-cannot assume that throughput advice from one path apply to new paths.
+cannot assume that throughput advice from one path applies to new paths.
 Advice for the client-to-server direction and the server-to-client direction of
 each path are independent, and are expected to be different, for reasons including
 asymmetric link capacity and path diversity.
@@ -268,7 +268,7 @@ as determined by congestion control.
 Endpoints that receive this signal therefore need to treat the information as advisory.
 
 The fact that an endpoint requests throughput advice does not necessarily mean
-that it will adhere to them; in some cases, the endpoint cannot. For
+that it will adhere to advice; in some cases, the endpoint cannot. For
 example, a flow could initially be used to serve video chunks, with the client
 selecting appropriate chunks based on received advice, but later switch to a
 bulk download for which bitrate adaptation that cannot be similarly controlled. Composite flows


draft-ietf-scone-protocol.md
```diff
@@ -193,9 +194,12 @@ are expected to last for more than one monitoring period; see {{time}}.
 
 ## Unspecified Scope
 
-Modifying a packet does not prove that the throughput that is indicated would be
-achievable.  A signal that is sent for a specific flow is likely enforced at a
-different scope.  The extent of that scope is not carried in the signal.
+Modifying a packet does not prove that the throughput that is indicated
+would be achievable.
+A signal that is sent for a specific flow
+is likely to apply to a collection of flows,
```
```suggestion
could apply to a collection of flows,
```

draft-ietf-scone-protocol.md
```diff
@@ -495,12 +500,10 @@ which necessarily includes congestion control
 and might include other, application-specific constraints.
 
 The relatively long duration of the monitoring period
-means that this is preferable to disabling sending completely.
+means that allowing advice to expire is preferable
+to disabling sending completely.
```
afrind: It's not clear why "disabling sending completely" is the alternative to allowing advice to expire.  Isn't another alternative the previously discussed (and rejected) idea of sticky advice?

draft-ietf-scone-protocol.md
```diff
@@ -934,9 +937,13 @@ with an implementation that fully conforms to the specification of {{packet}}.
...
 exchanged in other protocols.
-This could result in damage to those protocols.
+If the first five bytes of the datagram match
+the QUIC long header byte and SCONE version,
+the network element might modify the signal,
+resulting in damage to those protocols.
 
-The most serious damage occurs when every datagram is modified,
+The most serious damage occurs when every datagram matches
+and is subsequently modified,
```
afrind: Are there any other possible mitigations, like remembering flows a receiver indicated it's willing to receive SCONE packets on?

afrind (approved):
Thanks for the quick turnaround.  I have a few small comments.

draft-ietf-scone-protocol.md
martinthomson: Yeah, a rework is in order here.

draft-ietf-scone-protocol.md
```diff
@@ -193,9 +194,12 @@ are expected to last for more than one monitoring period; see {{time}}.
...
 
-Modifying a packet does not prove that the throughput that is indicated would be
-achievable.  A signal that is sent for a specific flow is likely enforced at a
-different scope.  The extent of that scope is not carried in the signal.
+Modifying a packet does not prove that the throughput that is indicated
+would be achievable.
+A signal that is sent for a specific flow
+could apply to a collection of flows,
+rather than a single flow.
+The scope of the flows that are included not carried in the signal.
```
```suggestion
The scope of the flows that are included is not carried in the signal.
```
Is this the intent?

kazuho (approved):

draft-ietf-scone-protocol.md
mjoras:
```suggestion
The scope of the flows that are included is not carried in the signal.
```
One more change.

martinthomson: Kazuho caught that already :)

PR #123, opened 2026-01-07
Use singular verb and pronoun forms consistently for "throughput advice"

afrind:
The AIs told me:

"Advice" is an uncountable noun, so it's always treated as singular, taking singular verbs (e.g., "advice is helpful") and pronouns (e.g., "give it to them").


martinthomson (approved):
Thanks.  Not as bad as I feared.

Looking for an RPC job?

PR #124, opened 2026-01-07
Clarify SCONE support indication placement in QUIC

afrind:


draft-ietf-scone-protocol.md
```diff
@@ -555,7 +555,7 @@ https://martinthomson.github.io/quic-pick/#seed=draft-ietf-scone-protocol-indica
 A client that uses a QUIC version that includes length-delimited packets,
 which includes QUIC versions 1 {{QUIC}} and 2 {{!QUICv2=RFC9369}},
 can include an indicator of SCONE support
-at the end of datagrams that start a flow.
+following the QUIC packets in datagrams that start a flow.
```
martinthomson: I think that it is genuinely "at the end", not following.

As in [packet][packet][whatever][indication], not [packet][packet][indication][whatever].  Recognizing that [packet][packet][indication] might be what you are thinking.

afrind: Ok, I wanted to capture that it's after the length delimited packets more explicitly.

Something other than [packet][packet+indication]

Maybe

```suggestion
outside of the QUIC packets at the end of datagrams that start a flow.
```

But I don't love it either.

martinthomson (approved):

PR #128, opened 2026-02-04
Trim out redundant congestion control text

martinthomson:
I've taken a couple of tiny things for the introductory material, but mostly this was just redundant.


kazuho (approved):
Thank you for doing this. I have some suggestions, but I'd be fine with what we already have.

huitema (approved):
I liked my original text, but I agree that this PR reflects the consensus from WG discussions.

Please apply Kazuho's suggestions.

martinthomson:
To be clear, I also liked your original text.  But it ended up being somewhat redundant. So it goes.

PR #133, opened 2026-02-19
editorial review

mirjak:


draft-ietf-scone-protocol.md
```diff
@@ -383,7 +382,8 @@ Source Connection ID field, which is the case for packets with a short header
 ({{Section 5.2 of INVARIANTS}}), the Source Connection ID field is empty
 and the Source Connection ID Length field is set to 0.
 
-SCONE packets MUST be included as the first packet in a datagram.
+SCONE packets are always coalesced with other QUIC packets in the same datagram and
```
mirjak: I added this here because this is mentioned later but it seems important to already make this clear here.

draft-ietf-scone-protocol.md
```diff
@@ -142,19 +142,19 @@ version, process the QUIC packet, and then record the indicated rate.
 Throughput advice only applies to the direction and path for which it is
 received.  A connection that migrates or uses multipath
 {{?QUIC-MP=I-D.ietf-quic-multipath}}
-cannot assume that throughput advice from one path applies to new paths.
-Advice for the client-to-server direction and the server-to-client direction of
+cannot assume that the throughput advice from one path applies to other paths.
```
martinthomson: "advice" is an uncountable noun and so it doesn't require the use of an article like "the" unless the text in question is referring to a specific piece of advice.  The changes you've suggested throughout do not fit that pattern.

In this specific case, you might use "the", but you don't have to.  To use "the", it might be better to say "... cannot assume that the throughput advice received for a specific network path applies to any other path."  That's more words and no more correct.


```suggestion
cannot assume that throughput advice from one path applies to other paths.
```

draft-ietf-scone-protocol.md
```diff
@@ -142,19 +142,19 @@ version, process the QUIC packet, and then record the indicated rate.
...
+of each path as they see it fit.
 
 
 # Applicability
 
-This protocol only works for flows that use the SCONE packet ({{packet}}).
+This protocol only works for flows that use QUIC {{QUIC}} (see also {{packet}}).
 
-The protocol requires that packets are modified as they transit a
+The protocol requires that SCONE packets are modified as they transit a
```
martinthomson: The change is correct in a sense, but you are changing the meaning of the statement.  The statement was to note the general capability that is expected to be present, it is not a description of how SCONE functions.

martinthomson (commented):
There are a few good changes here, but it's a bit of a challenge picking them out from the rest.  Would it be too much to ask for more targeted changes?

draft-ietf-scone-protocol.md
```diff
@@ -142,19 +142,19 @@ version, process the QUIC packet, and then record the indicated rate.
...
 asymmetric link capacity and path diversity.
 Applications can use SCONE in either or both directions
-of each path as they see fit.
+of each path as they see it fit.
 
 
 # Applicability
 
-This protocol only works for flows that use the SCONE packet ({{packet}}).
+This protocol only works for flows that use QUIC {{QUIC}} (see also {{packet}}).
```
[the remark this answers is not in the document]
mirjak: The point is that just saying in order to use a protocol you have to use the respective packets seems meaningless to me. So I wasn't sure what this sentence is suppose to tell me. However, the fact that you can use SCONE only together with an e2e quic connection should be stated clearly and right at the beginning of a applicability section and is actually missing as an explicit statement throughout the whole document.

mirjak:
> There are a few good changes here, but it's a bit of a challenge picking them out from the rest. Would it be too much to ask for more targeted changes?

I happy to do something else but you need to tell me what. However, you are the authors, so please feel free to edit this PR directly in anyway you want. I see these changes (except the one about use of QUIC; however, the original sentence was/is unclear to me) as editorial, so I don't have any strong opinions here. Just trying to help as I was reviewing the draft yesterday.

ihlar (commented):
A number of good changes here, some that could do with a bit of tweaking. 
But agree with @martinthomson that the changes from indefinite to definite articles are not really improving readability.

PR #138, opened 2026-03-12
Cherry pick some of the good stuff from #133

martinthomson:
There are plenty of good changes, but also a few things that could be better.

Closes #133.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -147,15 +147,17 @@ Advice for the client-to-server direction and the server-to-client direction of
 each path are independent, and are expected to be different, for reasons including
 asymmetric link capacity and path diversity.
 Applications can use SCONE in either or both directions
-of each path as they see fit.
+of each path at the discretion of endpoints.
 
 
 # Applicability
 
-This protocol only works for flows that use the SCONE packet ({{packet}}).
+This protocol can provide throughput advice only for QUIC flows
+where endpoints send SCONE packets ({{packet}}).
 
-The protocol requires that packets are modified as they transit a
-network element, which provides endpoints strong evidence that the network
+The operation of the SCONE protocol depends on network elements
+that are able to modify packets as they are forwarded.
+This provides endpoints strong evidence that the network
 element has the power to apply a rate limiting policy; though see {{security}} for
 potential limitations on this.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-protocol.md
```diff
@@ -506,10 +511,10 @@ applications MUST apply the lowest throughput advice
 received during any monitoring period; see {{time}}.
 
 After a monitoring period ({{time}})
-without receiving throughput advice,
-any previous advice expires.
+without receiving any throughput advice,
+the previous advice expires.
 Endpoints can remove any constraints
-placed on throughput based on receiving throughput advice.
+placed on throughput based on received throughput advice.
```
huitema: The change is fine, but the paragraph is a bit weird. We want to say that after the advice expires, the endpoint can loosen the brakes. That's fine, but there is no linkage between the two sentences. Should there be?

huitema (commented):
I generally like the update but I have a few comments.

draft-ietf-scone-protocol.md
huitema:
```suggestion
that resulted from the expired throughput advice.
```

draft-ietf-scone-protocol.md
```diff
@@ -252,11 +254,11 @@ Rate limiting policies often apply on the level of a device or subscription,
...
 
-Network elements that apply throughput advice to a flow
-that provides an encrypted tunnel for an encapsulated flow
-(such as {{?CONNECT-UDP=RFC9298}})
-only applies to the outermost flow.
-Advice can be applied on flows that are subsequently encapsulated,
+When throughput advice is applied to a QUIC flow
+that encapsulates tunneled flows (such as {{?CONNECT-UDP=RFC9298}})
+network elements can only apply the advice to the outermost flow.
+Advice can be applied to packets in flows
```
kazuho: Subject of the 1st sentence is network elements, while that of the 2nd is endpoints. To highlight the contrast, it'd be better to make that explicit and use different verbs.

```suggestion
When network elements provide throughput advice to a QUIC flow
that encapsulates tunneled flows (such as {{?CONNECT-UDP=RFC9298}})
they can only provide the advice to the outermost flow.
Endpoints can apply the throughput advice to packets in flows
```

kazuho (approved):
LG overall.

kazuho (approved):

PR #142, opened 2026-03-15
streamlining "achievable" and "sustainable"

zaheduzzaman:
We are using both of the terms, changes sustainable to be inline with the abstract

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -84,9 +84,9 @@ adapting their send rate.
 The Standard Communication with Network Elements (SCONE) protocol
 is negotiated by QUIC endpoints.
 SCONE provides a means for a network to signal its present best estimate
-for maximum sustainable throughput,
+for maximum achievable throughput,
 or throughput advice,
-associated with the flows of UDP datagrams that QUIC exchanges.
+associated with the flows of UDP datagrams that QUIC exchanges by the endpoits.
 
 Any network function that is able to update the content of UDP datagrams
 qualifies as a network element that can use SCONE packets
@@ -116,7 +116,7 @@ which are always coalesced with ordinary QUIC packets that they send.
 
 Networks that have rate limiting policies can detect flows that include
 SCONE packets.  The network, via an on-path network element, can indicate a maximum
-sustainable throughput by modifying the SCONE packet as it transits the
+achievable throughput by modifying the SCONE packet as it transits the
 network element.
 
 The propagation of SCONE packets, including the throughput advice that is added,
@@ -128,7 +128,7 @@ is shown in {{f-scone}}.
 | Sender |    | Element |     | Receiver |
 +---+----+    +----+----+     +----+-----+
     |              |               |
-    +--- SCONE --->|  SCONE+advice |
+    +--- SCONE --->| SCONE(advice) |
     |    +QUIC     +---- +QUIC --->|
     |              |               |  Validate QUIC packet
     |              |               |  and record advice
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -86,7 +86,7 @@ is negotiated by QUIC endpoints.
 SCONE provides a means for a network to signal its present best estimate
 for maximum achievable throughput,
 or throughput advice,
-associated with the flows of UDP datagrams that QUIC exchanges by the endpoits.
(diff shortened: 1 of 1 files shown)


kazuho:
> We are using both of the terms, changes sustainable to be inline with the abstract

Nice catch. I think the right fix is to change "achievable" in the abstract to "sustainable".

In the body, after we introduce the monitoring period, using "achievable" can be fine in context. But in the abstract, "achievable" without that context could suggest the throughput of short bursts, whereas SCONE advice is intended to describe a rate that is sustainable over the monitoring period.

Using "sustainable" in the abstract avoids that ambiguity and is more consistent with the introduction.

zaheduzzaman:
What you really want to say here is that the "throughout advice" is a achievable rate by the endpoints during the monitoring period that the network *thinks* it can sustained or based on a policy for that particular UDP flow. 

With that I think sustainable is over used in the document as the network may not/can not guarantee the advice, it just gives its best estimation. I would say we don't use the "maximum sustainable throughput" in the document, rather say "maximum achievable throughput by the endpoints that the network could sustain for the {time} period".   

Will that work?

draft-ietf-scone-protocol.md
```diff
@@ -84,9 +84,9 @@ adapting their send rate.
 The Standard Communication with Network Elements (SCONE) protocol
 is negotiated by QUIC endpoints.
 SCONE provides a means for a network to signal its present best estimate
-for maximum sustainable throughput,
+for maximum achievable throughput,
```
huitema, line 87: I think achievable does not capture what SCONE does. If clients want to find out what is achievable, they can do that by just testing various rates until they get congestion feedback. The goal really is to have a duration element, and I think sustainable is a better word for that.

draft-ietf-scone-protocol.md
```diff
@@ -84,9 +84,9 @@ adapting their send rate.
...
 or throughput advice,
-associated with the flows of UDP datagrams that QUIC exchanges.
+associated with the flows of UDP datagrams that QUIC exchanges by the endpoits.
```
huitema: I don't think that's right. If you want to be pedantic, you could say:

```suggestion
associated with the flows of UDP datagrams that the endpoints exchange using QUIC.
```

But I don't think that's needed.

draft-ietf-scone-protocol.md
```diff
@@ -116,7 +116,7 @@ which are always coalesced with ordinary QUIC packets that they send.
 
 Networks that have rate limiting policies can detect flows that include
 SCONE packets.  The network, via an on-path network element, can indicate a maximum
-sustainable throughput by modifying the SCONE packet as it transits the
+achievable throughput by modifying the SCONE packet as it transits the
```
huitema, line 119: Please don't make that change.

huitema (changes requested):
I don't think that PR is need or particularly useful. In any case, no, saying "achievable" instead of "sustainable" is not a clarification.

ihlar:
While neither term is perfect, I think "sustainable" is the right one to use here. As others have mentioned, "achievable" can be interpreted in a way that limits instantaneous burst rates, which would be an unfortunate consequence, in my opinion.

martinthomson (changes requested):
My feelings on this are not as strong as @huitema's, but I'm not seeing a distinct improvement.

There is a point of tension between the expectations regarding the word sustainable (which has now picked up connotations that are unhelpful in this context) and the word achievable.  What is nice about the existing text is that "sustainable throughput" can be read to mean "achievable, sustained throughput".  That is, in the sense that flows are expected to be able to reach the target throughput and stay there.

Not including some mention of sustained throughput is going to lead to the sorts of problems Christian mentions.

draft-ietf-scone-protocol.md
zaheduzzaman: OK. This is inline with what I wrote - "maximum achievable throughput by the endpoints that the network could sustain for the {time} period".

zaheduzzaman:
I can agree both sustainable and acheivable is not doing their job to explain the expectations. We are having this discussion now couple of times. 

I would then suggest what I had suggested before, lets describe what we mean by maximum sustainable throuput and move on with it. Something like this would be helpful I think - "maximum achievable throughput by the endpoints that the network could sustain for the {time} period". And update the abstract to use "sustainable".

PR #144, opened 2026-03-20
Only allow one SCONE packet per datagram

### LINKED ISSUE #143: Can we limit to one SCONE packet?
There is no reason to include more than one SCONE packet, so can we prohibit the inclusion of more than one?  This came up during implementation.  We had an implementation bug as a result of accepting many SCONE packets (I'd implemented using recursion and a large packet full of SCONE was blowing the stack).  Given that there is no need for more than one SCONE packet, can we make that a requirement?

martinthomson:
But the receiver behavior is only discarding.

Closes #143.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -394,6 +394,8 @@ in network elements.
 This is also necessary in many cases for QUIC versions 1 and 2
 because packets with a short header cannot precede any other packets.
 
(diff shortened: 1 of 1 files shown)


britram:
this makes sense to me, and i'm surprised this wasn't already here. ship it (as individual)

huitema (approved):
Yes, that's what we agreed.

ihlar (approved):

kazuho (approved):

PR #147, opened 2026-03-31
Ah, the spin bit

### LINKED ISSUE #146: Should SCONE mention the Spin Bit?
The spin bit is only on 1-RTT packets, so there isn't actually a conflict between the bit and the fields in SCONE.

But, whatever is watching the spin bit has to either ignore a SCONE packet or parse ahead to the 1-RTT header. I imagine they'll do the former, but I wonder if we should add some text about this.

I have not implemented a spin bit observer; I imagine they just ignore long-header datagrams, so maybe this flows naturally?

martinthomson:
Follow through on Christian's suggestion

Closes #146.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -892,6 +892,27 @@ network element applied throughput advice.
 Senders will therefore proceed as though there was no advice.
 
 
+## Interactions With The Latency Spin Bit {#spin}
+
+Endpoints that observe the recommendations for setting the latency spin bit,
+as defined in {{Section 17.4 of QUIC}},
+generate two competing signals for elements on the network path.
+
+A QUIC packet that is coalesced with a SCONE packet
+might make it harder for network elements to detect the latency spin bit,
+because the SCONE packet moves the bit from the first byte of the datagram
+to after the SCONE packet.
+
+Network elements that ignore datagrams with a SCONE packet
+could miss changes in the spin bit.
+This could introduce measurement error
+that inflates round-trip time estimates.
+
+Endpoints that enable the spin bit can avoid this issue
+by deferring the inclusion of SCONE packets
+when a spin bit edge is being communicated in a QUIC packet.
+
+
 # Security Considerations {#security}
 
 The modification of packets provides endpoints proof that a network element is


kazuho (approved):

huitema (approved):
That works, of course. The only question is whether we really need to say something about the spin bit, but, why not?

draft-ietf-scone-protocol.md
```diff
@@ -892,6 +892,27 @@ network element applied throughput advice.
...
+
+A QUIC packet that is coalesced with a SCONE packet
+might make it harder for network elements to detect the latency spin bit,
+because the SCONE packet moves the bit from the first byte of the datagram
+to after the SCONE packet.
+
+Network elements that ignore datagrams with a SCONE packet
+could miss changes in the spin bit.
+This could introduce measurement error
+that inflates round-trip time estimates.
```
mirjak, line 909: Proposal to add a bit of quantification as usually this inflation should be small:
```suggestion
that inflates round-trip time estimates depending on when the next packet arrives.
```

martinthomson: I don't think we need to address that.  It's also possible that the non-SCONE-bearing datagram could be reordered ahead of the SCONE-bearing datagram.   (I also question that assertion about "small", because we don't know when the next datagram will have been sent.)

mirjak: Actually I don't know if we really need this at all. This is no different than losing the packet that changes the spin bit.

martinthomson: It is different: in some implementations, it would be reasonable to expect the first packet after a period of quiescence to include SCONE, which also interferes with the spin bit.  If you would prefer we not say anything about the spin bit, I can understand that; we can close this pull request without merging.  But this also seems harmless.

huitema: We may well be better off by not mentioning the spin bit at all. Do we have actual deployment of network management tools relying on the spin bit? Do we know how hard it would be for such tools to just jump over the SCONE header?

martinthomson: Hmm, given where we are, it might pay to just poll the working group.  If anyone cares, we can add this text.  If they don't, we close the PR.  Then it becomes a chair problem and that's the sort of problem I like (any problem not mine, that is).

huitema: +1.

mirjak: Even if after an idle time the spin bit changes and there is a scone packet, I don't think this is a real issue as the next packet will have the spin bit. I agree that maybe there is not much special to say here as there can be may reason for missing packet with the bit change and you anyway have to account for that in your measurement. In the end you can only measure, what you can measure...

PR #150, opened 2026-04-12
Grammatical error

martinthomson:
Stuart pointed at this text and it doesn't read well.

This does nothing to address the concern raised in that email, but I thought it worth getting the language fixed anyway.


draft-ietf-scone-protocol.md
```diff
@@ -291,7 +291,7 @@ The fact that an endpoint requests throughput advice does not necessarily mean
 that it will adhere to advice; in some cases, the endpoint cannot. For
 example, a flow could initially be used to serve video chunks, with the client
 selecting appropriate chunks based on received advice, but later switch to a
```
kazuho: It could make sense to talk about adaptive bitrate selection while we describe video.

```suggestion
selecting chunks of different bitrates based on received advice, but later switch to a
```

kazuho (approved):

PR #152, opened 2026-04-30
Avoid Indicator Ossification

### LINKED ISSUE #151: Prevent the SCONE indicator from being an ossification vector
Interpreted literally, I believe the only logic the network element needs for the SCONE Indicator is:

`if (udp.payload.length >= 2 && last_two_bytes() == kSconeIndicator)
      FivetupleIsScone(fivetuple)`

In some conversations, I'm hearing that the network elements might be doing a little more than this. This is a potential ossification vector; for example, if it's parsing long headers to get packet lengths and find where the indicator should be, a QUICv1 parser will fail on v2 because the long header type codepoints are different. Even if they implement both versions, if we roll out QUICv3, then it will be hard to deploy because "you can't get SCONE with it".

It would be a good idea to specifically point network elements to [INVARIANTS]. If they must apply filters beyond the code snippet above, they could sensibly look for the long header bit and at least 9 bytes. This is not in the invariants, but it would also be reasonable to screen out datagrams < 1200 bytes, as I have trouble imagining that future versions of QUIC will not have these sorts of anti-amplification measures.

Elements might also ignore any packet after the first in the 5-tuple, if they want.

I would prefer MUST NOTs here, but at the very least it would be helpful to show how to avoid accidental ossification.

martinduke:
Fixes #151 

I'm unsure about the details and am happy to entertain improvements.

### DIFF
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -623,6 +623,30 @@ Despite these limitations,
 having an indication might allow network elements to change their starting posture
 with respect to their enforcement of their rate limit policies.
 
+## Avoiding Ossification When Reading the Indicator
+
+A network element could classify all 5-tuples where the first observed UDP
+datagram ends in the indicator bytes as potential SCONE. A network element MAY
+apply further criteria to further reduce this set and minimize false positives.
+However, it SHOULD NOT apply criteria that reduce the ability of new QUIC
+versions to employ SCONE. In particular, such criteria should consult the QUIC
+version invariants in {{INVARIANTS}}.
+
+For example, a network element could observe that the packet has the signature
+of a QUIC version 1 or 2 packet, including use of a long header, and if so,
+check that it meets the first-flight minimum datagram length of 1200 bytes for
+that version. However, if the datagram matches a QUIC long header with an
+unknown version, and has the indicator, it SHOULD treat it as a potential SCONE
+flow.
+
+In principle, a new QUIC version could have length-delimited short header
+packets and allow them in the first flight of a connection. However, these
+packets would not contain an explicit QUIC version number. Furthermore, short
+headers can be the first packets in a QUICv1 or QUICv2 5-tuple after an address
+or port migration, and these cannot contain the indicator. Therefore, network
+elements MAY ignore short header packets even though it is theoretically
+possible for future QUIC versions to support the SCONE indicator in a short
+header.
 
 ## Indications for Migrated Flows
 
--- a/draft-ietf-scone-protocol.md
+++ b/draft-ietf-scone-protocol.md
@@ -623,31 +623,6 @@ Despite these limitations,
 having an indication might allow network elements to change their starting posture
 with respect to their enforcement of their rate limit policies.
 
-## Avoiding Ossification When Reading the Indicator
-
-A network element could classify all 5-tuples where the first observed UDP
-datagram ends in the indicator bytes as potential SCONE. A network element MAY
-apply further criteria to further reduce this set and minimize false positives.
-However, it SHOULD NOT apply criteria that reduce the ability of new QUIC
-versions to employ SCONE. In particular, such criteria should consult the QUIC
-version invariants in {{INVARIANTS}}.
-
-For example, a network element could observe that the packet has the signature
-of a QUIC version 1 or 2 packet, including use of a long header, and if so,
-check that it meets the first-flight minimum datagram length of 1200 bytes for
-that version. However, if the datagram matches a QUIC long header with an
-unknown version, and has the indicator, it SHOULD treat it as a potential SCONE
-flow.
-
-In principle, a new QUIC version could have length-delimited short header
-packets and allow them in the first flight of a connection. However, these
-packets would not contain an explicit QUIC version number. Furthermore, short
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-protocol.md
```diff
@@ -643,6 +642,31 @@ any time they send a QUIC PATH_CHALLENGE or PATH_RESPONSE frame.
...
 
+## Avoiding Ossification When Reading the Indicator
+
+A network element could classify all 5-tuples where the first observed UDP
+datagram ends in the indicator bytes as potential SCONE. A network element MAY
+apply further criteria to further reduce this set and minimize false positives.
+However, it SHOULD NOT apply criteria that reduce the ability of new QUIC
+versions to employ SCONE. In particular, such criteria should consult the QUIC
+version invariants in {{INVARIANTS}}.
+
```
martinthomson: Why not stop here?

draft-ietf-scone-protocol.md
```diff
@@ -643,6 +642,31 @@ any time they send a QUIC PATH_CHALLENGE or PATH_RESPONSE frame.
 This applies to both client and server endpoints,
 but only if the peer has sent the transport parameter; see {{tp}}.
 
+## Avoiding Ossification When Reading the Indicator
+
+A network element could classify all 5-tuples where the first observed UDP
+datagram ends in the indicator bytes as potential SCONE. A network element MAY
+apply further criteria to further reduce this set and minimize false positives.
+However, it SHOULD NOT apply criteria that reduce the ability of new QUIC
+versions to employ SCONE. In particular, such criteria should consult the QUIC
```
```suggestion
versions to employ SCONE. SCONE operates independently of any
specific QUIC version, so any criteria needs to fit within the QUIC
```

draft-ietf-scone-protocol.md
```diff
@@ -643,6 +642,31 @@ any time they send a QUIC PATH_CHALLENGE or PATH_RESPONSE frame.
...
+
+In principle, a new QUIC version could have length-delimited short header
+packets and allow them in the first flight of a connection. However, these
+packets would not contain an explicit QUIC version number. Furthermore, short
+headers can be the first packets in a QUICv1 or QUICv2 5-tuple after an address
+or port migration, and these cannot contain the indicator. Therefore, network
+elements MAY ignore short header packets even though it is theoretically
+possible for future QUIC versions to support the SCONE indicator in a short
+header.
+
```
huitema: I agree with Martin T. We can do without these last 2 paragraphs. 

If we really want to belabor the point, it would be simpler to add a single sentence ad the end of the first paragraph, after `version invariants in {{INVARIANTS}}.` Something like, "In particular, the version number in the first observed packet will not always match QUICv1, and the length of the that packet could be lower than the minimal value set by QUICv1." But I don't think that's necessary.

huitema (commented):
I think we should keep the text short.

draft-ietf-scone-protocol.md
martinduke: Done

draft-ietf-scone-protocol.md
```diff
@@ -643,6 +642,15 @@ any time they send a QUIC PATH_CHALLENGE or PATH_RESPONSE frame.
 This applies to both client and server endpoints,
 but only if the peer has sent the transport parameter; see {{tp}}.
 
+## Avoiding Ossification When Reading the Indicator
+
+A network element could classify all 5-tuples where the first observed UDP
+datagram ends in the indicator bytes as potential SCONE. A network element MAY
+apply further criteria to further reduce this set and minimize false positives.
+However, it SHOULD NOT apply criteria that reduce the ability of new QUIC
+versions to employ SCONE. SCONE operated independently of any specific QUIC
```
```suggestion
versions to employ SCONE. SCONE operates independently of any specific QUIC
```
Fix the typo.

huitema (approved):
Looks good, except for a typo.

draft-ietf-scone-protocol.md
```diff
@@ -643,6 +642,15 @@ any time they send a QUIC PATH_CHALLENGE or PATH_RESPONSE frame.
 This applies to both client and server endpoints,
 but only if the peer has sent the transport parameter; see {{tp}}.
 
+## Avoiding Ossification When Reading the Indicator
+
+A network element could classify all 5-tuples where the first observed UDP
+datagram ends in the indicator bytes as potential SCONE. A network element MAY
+apply further criteria to further reduce this set and minimize false positives.
```
```suggestion
apply further criteria to further reduce the set of flows that are SCONE candidates and minimize false positives.
```

martinthomson (approved):
"this set" wasn't quite direct enough, but this is fine either way.

draft-ietf-scone-protocol.md
zaheduzzaman: "SCONE candidates" seems to be something new and not defined,  and to avoid two "further"

I would rewrite as

Apply additional criteria to further reduce the set of flows that support SCONE and minimize the false positives.

martinthomson: Like so, I guess...
image


```suggestion
apply additional criteria to further reduce the set flows that are identified
as potentially supporting SCONE, reducing the likelihood of false positives.
```

zaheduzzaman: yap!! much better.

PR #159, opened 2026-08-11
Gorry's AD review

martinthomson:
The thing is, any document at any stage, when subject to additional scrutiny, can be improved.  Hopefully, these are all improvements and not just shuffling things around.

Nothing more than editorial changes here, by my reckoning.


draft-ietf-scone-protocol.md
```diff
@@ -923,21 +951,19 @@ the peer, with the result being that the endpoint receives a false
...
-The congestion controller employed by a sender provides
-real-time information about the rate at which the network path is delivering
-data.
+The capabilities required of an off-path attacker are substantially
+similar to those of on path elements.
+An off-path attacker can generate throughput advice
+that will be accepted by an endpoint
+if it has the ability to damage packets
+in a way that could be able to affect throughput capacity of the flow.
+The one exception is where an off-path attacker
```
~~~suggestion
The one exception is an off-path attacker
~~~

I think adding "where" at this point in the sentence is awkward.

huitema (commented):
Looks good, except for one nit. I agree that these are editorial comments. I also agree that the reminder of Gorry's review requires more discussion.

ihlar (approved):

draft-ietf-scone-protocol.md
```diff
@@ -205,8 +205,11 @@ are expected to last for more than one monitoring period; see {{time}}.
 
 ## Unspecified Scope
 
-Modifying a packet does not prove that the throughput that is indicated
-would be achievable.
+Modifying a packet is not proof that the indicated throughput is achievable.
```
gorryfair: I struggled to slowly read that and work out what it said. Could this even be something like:
"The ability to modify a SCONE packet as it transits a network element does not prove that a flow could achieve the indicated throughput"?

draft-ietf-scone-protocol.md
gorryfair:
```suggestion
and change sending behavior more quickly than SCONE allows.
```

gorryfair:
```suggestion
SCONE packets MUST be coalesced with other QUIC packets
```

gorryfair: The text says: “SCONE packets are always coalesced with other QUIC packets (see Section 12.2 of [QUIC]) and MUST be included as the first packet in a datagram”, which of course is true, and although this says coalescing is needed, and explain how to, I could not find place where this says “MUST be Coalesced”. Can this text be changed to:
“SCONE packets MUST be coalesced with other QUIC packets (see Section 12.2 of [QUIC]) and MUST be included as the first packet in a datagram”

gorryfair (commented):
This is very close to what I was hoping for.
Please see suggestions in this github review and also consider the following two topics:
---
1) I added a proposal to change make a normative requirement to coalesce the SCONE packet (see proposed change).
---
2) From previous email thread:
/A network element SHOULD base its monitoring on the maximum value that was configured to apply during the preceding two monitoring periods/
- I am unsure why this couldn’t be: //A network element performing monitoring  MUST use a monitoring  period at least….”
- If it needs to be kept as a SHOULD then I think we ought to explain why, and what are the implications of lower and higher, which it might already perhaps?
This is a SHOULD and it is already explained: This allows applications time to receive advice and adapt their sending rate. A tighter duration is possible if there is better information available.  For instance, if SCONE signals are sent often enough that you can be sure that advice reaches the application within a shorter period than one cycle, you could apply tighter timings once you account for the possibility of loss and so forth.  The algorithm here is, as mentioned, illustrative.
>> Sorry, I was not clear and alas, I still do not understand.
I think question is: can the the value be less than the maximum value that was configured to apply during the preceding monitoring period?
- My reading was it cannot be less, but I wanted to see if the text can explain this.

draft-ietf-scone-protocol.md
martinthomson: Edit: nevermind, the weird rendering caught me out.  I thought that this was suggesting a removal of the text.

draft-ietf-scone-protocol.md
martinthomson: How about:

```suggestion
Just because a network element can set throughput advice,
that does not prove that the flow can achieve that rate.
```

huitema (commented):
Reading Martin's proposal. I think that removing the SHOULD is great. It clearly signals that this section is only advisory. After all, one of the most frequent policy is to add a rate limiter to a broadband connection. If an ISP does that, then having a way to tell the limit to the clients using SCONE is great, but the ISP does not need to bother implementing any kind of per flow limitation.

The text contains the right kind of advice.

draft-ietf-scone-protocol.md
```diff
@@ -762,14 +774,17 @@ A network element can then monitor affected flows
 to determine whether the provided throughput advice
 was followed.
 
-A network element SHOULD base its monitoring
+The simplest monitoring approach bases monitoring
 on the maximum value that was configured to apply
```
mirjak: Configure where to do what?

draft-ietf-scone-protocol.md
```diff
@@ -762,14 +774,17 @@ A network element can then monitor affected flows
...
 during the preceding two monitoring periods.
+This allows an additional monitoring period
+to allow for senders providing SCONE packets
+that carry updated throughput advice to receivers reliably.
```
mirjak: I find this new wording a bit confusing. Maybe having even something more high-level is better:


```suggestion
This allows an additional monitoring period
to compensate for additional delays
as SCONE packets are not delivered reliably.
```

mirjak: Any maybe this would be a good place to add a reference to the manageability draft directly?

draft-ietf-scone-protocol.md
martinthomson:
```suggestion
on the maximum value
that the network element was configured to apply to SCONE packets
```

draft-ietf-scone-protocol.md
```diff
@@ -736,11 +745,14 @@ to determine whether applications are following advice.
...
 This section outlines a method
 that a network element could use
-to determine whether advice is being followed.
+to determine whether a flow exceeds the value from provided throughput advice.
 Network deployments that choose to monitor
 are free to follow any monitoring regime that suits their needs.
 
-This monitoring algorithm is guidance only.
+The approach to monitoring that is documented here
+is largely illustrative;
```
```suggestion
This documented approach to monitoring that is largely illustrative;
```
I think we might have a grammatical error here?

kazuho (approved):

draft-ietf-scone-protocol.md
martinthomson: I think your version adds a grammatical error.  The "that" can be removed.

gorryfair: I'd be happy with: "This documented approach to monitoring is largely illustrative;"

gorryfair:
Is this ready to merge?

draft-ietf-scone-protocol.md
martinthomson:
```suggestion
This documented approach to monitoring is largely illustrative;
```

PR #160, opened 2026-08-11
Cite appman draft

martinthomson:


gorryfair (approved):
For me, this captures what is missing!

gorryfair (approved):
I think this is a useful improvement, for little cost.

